PR #185, opened 2026-05-03
💥 Drop Python 3.11, adopt PEP 695 generic syntax

### LINKED ISSUE #65: Adopt PEP 695 type parameter syntax
## Context

`grelmicro/_backends.py` uses `T = TypeVar("T")` and `grelmicro/cache/cached.py` uses `P = ParamSpec("P")` / `R = TypeVar("R")`.

Python 3.12+ supports `type T = ...` and native `class Foo[T]:` syntax (PEP 695).

## Blocked by

Requires dropping Python 3.11 support (see #B-013) or bumping minimum to `>=3.12`.

## Files

- `grelmicro/_backends.py`
- `grelmicro/cache/cached.py`

## Roadmap

Low priority. Not in scope for 1.0 (see [Roadmap to 1.0](#124)). Internal-syntax cleanup, depends on the Python-floor decision in #66.

grelinfo:
First step of the 0.21.0 runtime-foundation release ([#161](https://github.com/grelinfo/grelmicro/issues/161)).

## Summary

- Bump `requires-python` from `>=3.11` to `>=3.12`. RHEL 9 (App Stream `python3.12`) and RHEL 10 (default) ship 3.12 with UBI images available, so enterprise users are covered.
- Adopt PEP 695 generic syntax across `_backends.py`, `_config.py`, `_types.py`, `health/_types.py`, `trace/_instrument.py`, `tests/task/conftest.py`.
- Bump `tool.ruff.target-version` to `py312` and the CI matrix to `["3.12","3.13","3.14"]`.
- `_json.py` stays on `TypeAlias` with `# noqa: UP040` until ty resolves recursive PEP 695 expansion.

Closes #66, closes #65.

## Test plan

- [x] `uv run ruff check .`
- [x] `uv run ty check`
- [x] `uv run pytest tests` (1431 passed)
- [x] All pre-commit hooks passed (incl. coverage)


grelinfo:
Thanks for the catch. `cache/cached.py` was indeed not migrated by ruff's autofix because the decorator-factory pattern doesn't have a clean PEP 695 form: making either `cached` or the inner `decorator` generic with `[**P, R]` either binds the inner decorator to the outer scope's type parameters (breaking per-decoration-site inference) or requires a Protocol-typed return that adds more complexity than it removes. Reverted to module-level `ParamSpec`/`TypeVar` with an inline comment explaining why, and updated the changelog to list both `cached.py` and `_json.py` as documented exceptions to the PEP 695 sweep. Pushed in 6cfff61.

Issue #201, opened 2026-05-07
📦 Epic: Grelmicro app object — retire module-level registries


grelinfo:
Tracking epic for the `Grelmicro` class work. See [docs/architecture/grelmicro.md](https://github.com/grelinfo/grelmicro/blob/main/docs/architecture/grelmicro.md) for the full design.

## Why

Module-level registries (`sync_backend_registry`, `cache_backend_registry`, ...) and the global `_ALL_REGISTRIES` cause:

1. Import-time mutation. Hostile to tree-shake (#189).
2. Test isolation requires `registry.reset()` fixtures.
3. Multiple grelmicro setups in one process is impossible.

## Target

A single `Grelmicro` class. Features registered with `micro.use(...)`. Per-task `ContextVar` for ambient lookup. Last breaking change before 1.0.

## Scope

Migration spans three minor releases (`0.22.0` → `0.24.0`).

- [ ] Decide open questions (sub-issues with label `area:core`)
- [ ] 0.22.0 — land new API alongside old
- [ ] 0.23.0 — deprecate module-level helpers
- [ ] 0.24.0 — remove `_ALL_REGISTRIES` and free `lifespan()`
- [ ] 1.0.0 — `Grelmicro` is the only API

Sub-issues are tracked under the `area:core` label.

grelinfo:
All three remaining open rows from the Status table are shipped and verified against the codebase:

- **`Backend` impl → `Adapter` rename** — 0.22.0 changelog, every concrete class is `*Adapter` (`MemorySyncAdapter`, `RedisSyncAdapter`, `PostgresSyncAdapter`, `SQLiteSyncAdapter`, `KubernetesSyncAdapter`, `MemoryCacheAdapter`, `RedisCacheAdapter`, `MemoryRateLimiterAdapter`, `RedisRateLimiterAdapter`, `MemoryCircuitBreakerAdapter`). Protocols (`SyncBackend`, `CacheBackend`, `RateLimiterBackend`, `CircuitBreakerBackend`) kept.
- **`Provider` layer (was `Engine`)** — `grelmicro/providers/redis.py` and `grelmicro/providers/postgres.py` ship `RedisProvider` and `PostgresProvider`. Component-direct Provider API landed in #264.
- **Module-level registry removal (#207)** — `grelmicro.lifespan` and the `register`/`unregister`/`use_backend`/`use_registry` helpers are gone across the package. Patterns resolve via `Grelmicro.current()`.

[...]

PR #216, opened 2026-05-07
✨ Add Health module wrapping HealthRegistry

grelinfo:
## Summary

Implements #213 (Health module). Wraps a `HealthRegistry` and forwards the `check(...)` decorator and `run(...)` aggregator. Part of the #201 epic.

## Usage

```python
from grelmicro import Grelmicro
from grelmicro.health import Health

micro = Grelmicro(modules=[Health()])

@micro.health.check("redis")
async def redis_alive() -> None:
    ...

async with micro:
    report = await micro.health.run()
```

## Changes

- `grelmicro/health/_module.py`: `Health` class implementing `Module` (`kind="health"`)
- `grelmicro/health/__init__.py`: re-export `Health`
- `tests/health/test_module.py`: 10 tests
- `docs/changelog.md`: feature entry

## Test plan

- [x] All 10 new tests pass
- [x] Full pre-commit gate green
- [x] `micro.health` follows the locked default-vs-named resolution rule


grelinfo:
Superseded by the include() pattern PR. The Health module wrapper is the wrong direction - HealthRegistry is an entry point that should be registered explicitly via micro.include(), like FastAPI's app.include_router().

PR #221, opened 2026-05-08
💥 Rename TaskManager to Tasks (#218)

grelinfo:
## Summary

- Rename `TaskManager` to `Tasks`. Class still extends `TaskRouter`, mirroring FastAPI's `APIRouter` ← `FastAPI` shape.
- Move `grelmicro/task/manager.py` to `grelmicro/task/_tasks.py`. Move `tests/task/test_manager.py` to `tests/task/test_tasks.py`.
- Keep `from grelmicro.task import TaskManager` working via module-level `__getattr__` that emits `DeprecationWarning`. Removed in 1.0.0.
- Sweep docs, snippets, and README.

Closes #218.

## Test plan

- [x] `uv run pytest` (1494 passed)
- [x] `uv run ruff check grelmicro/ tests/`
- [x] `uv run ty check grelmicro/`
- [x] Coverage 100%
- [x] New tests cover the deprecation alias and the unknown-attribute path

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,12 +105,12 @@ from grelmicro.resilience import (
 from grelmicro.resilience.redis import RedisRateLimiterBackend
 from grelmicro.sync import LeaderElection, Lock
 from grelmicro.sync.redis import RedisSyncBackend
-from grelmicro.task import TaskManager
+from grelmicro.task import Tasks
 
 logger = logging.getLogger(__name__)
 
 # === grelmicro ===
-task = TaskManager()
+task = Tasks()
 sync.register(RedisSyncBackend("redis://localhost:6379/0"))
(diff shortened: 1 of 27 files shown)


grelinfo:
Addressed:
- Caching deprecated alias after first warn — done in b28abd5.
- Doc heading `## Task Manager` → `## Tasks` — done in b28abd5.

Intentionally not addressing the `__all__` comment: `TaskManager` is excluded by design so `from grelmicro.task import *` and IDE autocompletion only surface `Tasks`. Direct attribute access continues to work for back-compat. The alias is deprecated and slated for removal in 1.0.0; advertising it via `__all__` would push users back onto the legacy name.

PR #253, opened 2026-05-12
💥 Rename resilience Backend impls to *Adapter

grelinfo:
## Summary

- Rename three resilience backend impl classes to `*Adapter` to align with the locked architecture (#201): `MemoryRateLimiterBackend` → `MemoryRateLimiterAdapter`, `RedisRateLimiterBackend` → `RedisRateLimiterAdapter`, `MemoryCircuitBreakerBackend` → `MemoryCircuitBreakerAdapter`.
- The `RateLimiterBackend` and `CircuitBreakerBackend` Protocols stay as-is (Protocols are `*Backend`, impls are `*Adapter`).
- Clean cut, no deprecation shim (0.x policy). Update the unreleased changelog `### Breaking` entry to include the resilience classes.
- Sweep docs, snippets, README, and tests. Rename `_CountingRateLimiterBackend` test helper to `_CountingRateLimiterAdapter`.

Part of #201.

## Test plan

- [x] `uv run pytest` (1636 passed)
- [x] `uv run ruff check grelmicro/ tests/`
- [x] `uv run ty check grelmicro/`


grelinfo:
Addressed Copilot review:

**Applied** (commit d759bd5): swept "backend" → "adapter" in the docstrings of `RedisRateLimiterAdapter`, `MemoryRateLimiterAdapter`, and `MemoryCircuitBreakerAdapter` (class doc, `__init__`, `__aenter__`/`__aexit__`, and `Doc(...)` annotations). Module-header docstring updated too.

**Not applied** (out of scope): the positional-URL examples in `README.md` and `docs/index.md` (`RedisSyncAdapter("redis://...")`). Copilot is correct that those would `TypeError` against the current keyword-only constructors, but the issue pre-exists on `main` (introduced by #248 when the Redis constructors became keyword-only via `provider=` / `env_prefix=`). This rename PR only swaps the class name on those lines, so I'd rather fix the construction style in a focused follow-up rather than smuggling it into a rename sweep.

PR #254, opened 2026-05-12
📝 Fix README Redis adapter examples to use shared RedisProvider

grelinfo:
## Summary

The quickstart example in `README.md` (and the synced `docs/index.md`) constructed Redis adapters with positional URL arguments (`RedisSyncAdapter("redis://...")`). The constructors have been keyword-only since #248: positional `url=` no longer exists. Copy-pasting the snippet would raise `TypeError`.

Fix: build one `RedisProvider("redis://localhost:6379/0")` and pass it via `provider=` to all three adapters. This also showcases the implicit pool-sharing pattern that #248 introduced.

Surfaced during Copilot review on #253, deferred to keep that rename PR focused.

## Test plan

- [x] `uv run python -c "..."` constructs `RedisSyncAdapter`, `RedisCacheAdapter`, `RedisRateLimiterAdapter` with the new pattern without raising.
- [x] Pre-commit `readme-to-docs` hook regenerates `docs/index.md` from `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ from grelmicro.cache import JsonSerializer, TTLCache, cached
 from grelmicro.cache.redis import RedisCacheAdapter
 from grelmicro.health import HealthChecks
 from grelmicro.log import configure as configure_logging
+from grelmicro.providers.redis import RedisProvider
 from grelmicro.resilience import (
     CircuitBreaker,
     RateLimitExceededError,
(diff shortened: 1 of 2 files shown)


grelinfo:
Addressed Copilot review (commit 472007a): added the `redis` provider as the first item in `uses=[...]` so `Grelmicro` owns the lifecycle. The provider enters first, exits last (after the adapters that borrow it). Verified `_resolve_provider_sharing` leaves it alone (no `_owns_provider` attribute on the provider itself, so the loop skips it) and the explicit-`provider=` adapters still won't try to close it twice.

Note: the prior k3s testcontainer flake on Test Python 3.12 was unrelated to this docs change (the container exited mid-exec during the kubernetes sync backend integration test). Re-running CI.

PR #267, opened 2026-05-16
💥 Rename GCRAConfig to SlidingWindowConfig

grelinfo:
## Summary

Rename the user-facing rate-limiter algorithm config from `GCRAConfig` to `SlidingWindowConfig`, factory `RateLimiter.gcra(...)` to `RateLimiter.sliding_window(...)`, and discriminator value `"gcra"` to `"sliding_window"`. The module `grelmicro.resilience.algorithms.gcra` becomes `grelmicro.resilience.algorithms.sliding_window`.

The name was chosen after surveying how other rate-limiter libraries name this algorithm: throttled-py and ratelimit_meter ship `Leaky Bucket` and `GCRA` as **distinct** algorithms (queue vs meter formulations), so `LeakyBucketConfig` would mislead users coming from those libraries. `SlidingWindowConfig` matches the docstring (\"precise sliding window\"), the RFC 9211 `RateLimit-*` header model, and what API operators actually search for.

Internal strategy classes (`_RedisGCRA`, `_MemoryGCRA`, `_gcra_state`) keep their `GCRA` names since they describe the underlying algorithm and are private.

Pre-1.0 clean cut: no deprecation shim, no alias.

Closes #259.

## Test plan

- [x] `uv run pytest` (1625 passed)
- [x] `uv run ty check grelmicro`
- [x] `uv run ruff check grelmicro tests`


grelinfo:
@copilot Verdict: valid coverage gap. Added two tests in `4fcd04c`:

- `test_discriminator_values` — locks `TokenBucketConfig.type == "token_bucket"` and `SlidingWindowConfig.type == "sliding_window"`. A regression to `"gcra"` (or any other value) now fails fast.
- `test_rate_limiter_config_union_round_trips` — `TypeAdapter(RateLimiterConfig).validate_python({"type": "sliding_window", ...})` round-trips both arms of the union.

Tangentially: filed #268 to consider renaming the discriminator field from `type` to `kind` pre-1.0 (avoids the Python builtin shadow, matches Kubernetes/Argo/Tekton convention). Tagged `next`.

PR #276, opened 2026-05-18
✨ Restructure resilience into per-Pattern subpackages with shared CircuitBreaker

### LINKED ISSUE #163: ✨ Distributed CircuitBreaker state via CircuitBreakerBackend protocol
## Problem

Today `CircuitBreaker` keeps `state`, `consecutive_error_count`, `consecutive_success_count`, `_active_call_count`, and `_open_until_time` on the instance. Each replica trips its own breaker. A 50-replica deployment hitting a degraded downstream sees 49 replicas continue probing after the 50th opens. That weakens the central marketing bullet ("battle-tested in production at scale") and is the most common surprise readers report when comparing grelmicro to libraries like resilience4j.

## Required design

Mirror the shape of `RateLimiterBackend`:

```python
class CircuitBreakerBackend(Protocol):
    async def get_state(self, *, name: str) -> CircuitBreakerSharedState: ...
    async def transition(self, *, name: str, expected: ..., desired: ...) -> bool: ...
    async def increment_error(self, *, name: str, threshold: int) -> CircuitBreakerSharedState: ...
    async def increment_success(self, *, name: str, threshold: int) -> CircuitBreakerSharedState: ...
```

`CircuitBreaker` accepts an optional `backend=` (instance, registered name, or `None` for the existing local-only behavior, which stays the default). When set, `_try_acquire_call`, `_on_error`, and `_on_success` consult the backend with optimistic compare-and-set semantics. `last_error` / `last_error_time` stay local: per-replica diagnostics are useful and shipping them around hurts more than it helps.

Ship one Redis backend in `grelmicro/resilience/redis.py` (Lua scripts for the CAS path, mirroring how `RedisRateLimiterBackend` does its window math).

## Acceptance

- `CircuitBreakerBackend` protocol in `grelmicro/resilience/_protocol.py`.
- `RedisCircuitBreakerBackend` with full test coverage including: state convergence under concurrent errors from N tasks, half-open admission cap respected globally, manual `transition_to_*` calls visible to other replicas.
- Local-only mode unchanged for users that do not pass `backend=`.
- Reference doc explains the trade-off (one-line summary: "use shared state when you want one replica's circuit decision to short-circuit the rest; use local-only when each replica's downstream is independent").
- Comparison line in the resilience landing page so readers see we have it.
- Changelog entry.

## Out of scope

- Postgres backend (separate issue, lower priority): only file once the protocol stabilizes.
- Kubernetes-native breaker (no shared state primitive there).
- Sharing `last_error` across replicas.

Tracked under [#161](https://github.com/grelinfo/grelmicro/issues/161).

grelinfo:
## Summary

Three-commit refactor of `grelmicro.resilience`. Adds a fleet-wide Redis circuit breaker, unifies the local/shared paths into a single Strategy interface, and reorganises the package into per-Pattern subpackages with lazy imports.

- ✨ **`RedisCircuitBreakerAdapter`** — atomic Lua admission/record/transition, half-open admission cap enforced globally, `last_error_time` stays per-replica.
- ♻️ **CircuitBreakerBackend → bind/Strategy pattern** — every backend (memory and Redis) goes through `CircuitBreakerStrategy`. `register(breaker)` and the local fast-path are removed.
- 📦 **Per-Pattern subpackages** — `circuitbreaker/` and `ratelimiter/` each hold their algorithm configs, memory adapter, and Redis adapter. Top-level `grelmicro.resilience.algorithms`, `grelmicro.resilience.memory`, `grelmicro.resilience.redis` modules are gone.
- ✨ **PEP 562 lazy loader** — `from grelmicro.resilience import CircuitBreaker` no longer loads `RateLimiter` or any adapter. Same in the other direction.
- ✨ **Implicit memory adapter** — `CircuitBreaker("payments")` and `RateLimiter.token_bucket("api", capacity=10, refill_rate=1)` work without `Grelmicro(uses=[...])` setup. Fleet-wide opt-in stays explicit.
- ✨ **Factory classmethod** — `CircuitBreaker.consecutive_count(name, ...)` matches `RateLimiter.token_bucket` / `.sliding_window`. Algorithm-config module loads lazily on first call.
- 💥 **`CircuitBreakerConfig` is a `Discriminator("kind")` union** — instantiate `ConsecutiveCountConfig(...)` directly.
- 💥 **Component rename**: `Breaker` → `CircuitBreakers`, `RateLimit` → `RateLimiters`. Plural matches `Tasks`/`HealthChecks` convention.
- 💥 **`CircuitBreaker.__init__` drops kwargs API** — takes `(name, config=None, *, backend=None)` like `RateLimiter`. Env loading via `GREL_CIRCUIT_BREAKER_*` removed.
- 💥 Forced states (`FORCED_OPEN`, `FORCED_CLOSED`) no longer tick consecutive counters.

Closes #163.

## Test plan

- [x] `uv run pytest tests/` — 1626 / 1626 pass
- [x] `uv run ty check grelmicro/ tests/` — clean
- [x] Lazy-import boundary verified: `from grelmicro.resilience import CircuitBreaker` does not load `ratelimiter/`, and vice versa
- [x] Implicit memory adapter exercised via dedicated tests
- [x] Coverage at 100% under `pytest-unit + pytest-integration`
- [ ] First Copilot review pass
- [ ] Address Copilot comments


grelinfo:
Thanks for the review. All 9 comments evaluated and addressed in 677cef9.

## Doc-prose bugs (5)

The component rename sed used `\bBreaker\b → CircuitBreakers` with a word-boundary regex. It accidentally rewrote the prose phrase "Circuit Breaker" into "Circuit CircuitBreakers" in headings, docstrings, and intro sentences across 7 files. Reverted everywhere; the component class name plural lives only at the API surface, not in prose.

## HTTP header bugs (2)

Same regex hit `\bRateLimit\b → RateLimiters` and pluralized the RFC 9211 `RateLimit-*` and legacy `X-RateLimit-*` header names in docs and `_protocol.py`. Reverted. The wire-level header names are unchanged.

## Redis TTL bypass (3)

All three TTL comments are valid. The TTL-on-state-write pattern caused: (a) OPEN states silently expiring after `cool_down*2` with no traffic, then admitting the next call as CLOSED (skipping HALF_OPEN); (b) the same for HALF_OPEN-mid-probe; (c) manual transitions to non-OPEN states inheriting any stale TTL.

Fix: removed every `EXPIRE` from the Lua scripts and added a `PERSIST` in `_LUA_TRANSITION`. CB state is now sticky for the lifetime of the Redis store (matching resilience4j's process-lifetime semantics for in-JVM state). Storage cost is tiny per breaker name and bounded by the number of unique breakers. The global `half_open_capacity` cap is now correctly enforced even after long idle periods.

Tests: 1627/1627 pass.

PR #286, opened 2026-05-21
✨ Add Timeout resilience pattern (#176)

grelinfo:
## Summary

- Adds `Timeout`, a reconfigurable resilience pattern wrapping `asyncio.timeout`. `Timeout("db", seconds=2.0)` works as an async context manager (`async with db_timeout:`) and as a decorator on async functions.
- `TimeoutConfig` is a frozen three-paths Pydantic config with `seconds: PositiveFloat`. Env prefix `GREL_TIMEOUT_{NAME_UPPER}_`.
- Inherits `Reconfigurable[TimeoutConfig]`. Live deadline swaps leave in-flight scopes on the original deadline.
- Per-instance `ContextVar` stack so concurrent tasks and nested entries each get their own `asyncio.Timeout` scope.
- Sync functions rejected at decoration: asyncio cannot cancel sync code.

Closes #176.

## Scope cut from the original issue

\`policy: Literal["raise", "shield"]\` was in the issue sketch. First cut ships \`seconds\` only with raise semantics (the asyncio.timeout default). Shield-style "silent timeout" can land later if there is demand.

## Test plan

- [x] \`pytest tests/resilience/test_timeout.py\` 23 cases (config validation, three-paths construction, env path with name normalisation, success/timeout, concurrent reuse, nested entries, decorator sync rejection, reconfigure, in-flight scope keeps original deadline, outer-task cancellation propagates as \`CancelledError\`, composition under \`Retry\`)
- [x] \`pytest tests/\` full suite 1773 passed
- [x] \`ruff check\` + \`ruff format --check\`
- [x] \`ty check\`
- [x] \`mkdocs build --strict\`
- [ ] CI green
- [ ] Copilot review

### DIFF
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -6,7 +6,7 @@ The [roadmap](https://github.com/grelinfo/grelmicro/issues/124) and the [milesto
 
 ## Vocabulary
 
-- **Pattern**: user-facing class. `Lock`, `LeaderElection`, `TaskLock`, `TTLCache`, `RateLimiter`, `CircuitBreaker`, `Retry`, `Bulkhead`, `Fallback`.
+- **Pattern**: user-facing class. `Lock`, `LeaderElection`, `TaskLock`, `TTLCache`, `RateLimiter`, `CircuitBreaker`, `Retry`, `Bulkhead`, `Fallback`, `Timeout`.
 - **Adapter**: concrete implementation of a Backend Protocol. `RedisSyncAdapter`, `PostgresSyncAdapter`, `MemoryCacheAdapter`, `SQLiteSyncAdapter`, `KubernetesSyncAdapter`, and so on.
 - **Backend**: the Protocol class an Adapter satisfies. `SyncBackend`, `CacheBackend`, `RateLimiterBackend`, `CircuitBreakerBackend`.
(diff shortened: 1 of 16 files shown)


grelinfo:
Thanks @copilot. Verdicts:

**1. `Timeout.__aenter__`/`__aexit__` ContextVar inheritance (`grelmicro/resilience/timeout.py:112`)** — **Valid.** ContextVars copy into spawned tasks, so a child task created inside `async with timeout_policy:` inherits a stack containing the parent's `asyncio.Timeout` scope. Reference retention rather than a correctness bug (child push/pop only mutates the child's copy), but worth removing the footgun. Switched to a `dict[asyncio.Task, list[asyncio.Timeout]]` keyed by `asyncio.current_task()`. Child tasks now have a distinct identity and a fresh (empty) stack. Added a regression test `test_child_task_does_not_inherit_parent_scope`.

**2. `docs/snippets/resilience/timeout_environmental.py` missing `env_load=True`** — **Reject.** This matches the established codebase convention: `docs/snippets/resilience/fallback_environmental.py` and `docs/snippets/resilience/retry_environmental.py` both use the bare form without `env_load=True`. The `GREL_ENV_LOAD` opt-in is process-wide infrastructure documented once in `docs/architecture/config.md`, not duplicated on every Pattern's env snippet.

**3. `docs/resilience/timeout.md` Environmental section** — **Reject.** Same reason. Mirrors the Environmental sections of `docs/resilience/{retry,fallback}.md`.

PR #289, opened 2026-05-21
✨ Add Shield resilience pattern (#249)

### LINKED ISSUE #249: ✨ Add Shield: all-in-one outer-layer resilience primitive (timeout + retry + CUBIC adaptive throttling)
## Problem

`grelmicro.resilience` ships `Retry`, `RateLimiter`, and `CircuitBreaker` as separate, explicit primitives. Composing them correctly takes work: pick a backoff, pick a retry predicate, pick a rate-limiter algorithm, wire them in the right order, tune three sets of constants.

For the common case (\"wrap one call to one dependency, fail gracefully if the dependency goes bad\"), this is too many decisions. Users either don't compose them or compose them wrong. The library needs a one-line primitive that ships with sensible defaults and one knob for the SDK-specific exception types.

## Locked design

Full spec: [`docs/architecture/shield-spec.md`](../blob/main/docs/architecture/shield-spec.md).

`Shield` is the **outer layer of resilience**. Client libraries (requests, httpx, boto3, netmiko, gRPC) keep their own retry logic. `Shield` sits on top and adds a second, slower-timescale layer: per-attempt timeout, exponential-jittered retry gated by a consecutive-failure budget, and a CUBIC-style adaptive rate bucket that engages only after sustained slow-downs.

```python
from grelmicro.resilience import Shield

@Shield.api(timeout_errors=(httpx.TimeoutException, httpx.ConnectError))
async def fetch(url):
    return await client.get(url)
```

Three named profiles (`internal`, `api`, `slow`) bundle workload-dependent parameters; `api` is the default. The user picks a profile, passes the SDK's timeout exception types, and is done. Advanced users build a custom profile.

## API surface

```python
class Shield:
    @classmethod
    def internal(cls, name: str = ..., *, timeout_errors: tuple[type[BaseException], ...] = (TimeoutError, asyncio.TimeoutError), fallback: ... = None) -> Self: ...
    @classmethod
    def api(cls, name: str = ..., *, timeout_errors=..., fallback=...) -> Self: ...
    @classmethod
    def slow(cls, name: str = ..., *, timeout_errors=..., fallback=...) -> Self: ...

    @classmethod
    def from_config(cls, name: str, config: ShieldConfig) -> Self: ...

    async def run(self, fn, *args, **kwargs): ...
    def __call__(self, fn): ...                   # decorator form

# Module-level decorator for the anonymous case
@shield                                         # uses 'api' profile
async def fetch(...): ...

@shield.slow(timeout_errors=(...,))             # explicit profile
async def call_llm(...): ...
```

## Exception classification (API contract)

| Exception raised by wrapped call | Retried? | Triggers CUBIC shrink? | Consumes retry budget? |
|---|---|---|---|
| Any type in `timeout_errors` (or subclasses) | yes | yes | yes |
| Any other `Exception` subclass | **no — propagates immediately** | no | no |
| Subclass of `ResilienceError` | **no — propagates immediately** | no | no |
| `BaseException` outside `Exception` (cancellation, KeyboardInterrupt, SystemExit) | **no — propagates immediately** | no | no |

The user explicitly declares what \"transient\" means for their dependency by passing exception types to `timeout_errors=`. Everything else is treated as a hard failure and surfaces unchanged. The default `(TimeoutError, asyncio.TimeoutError)` covers `Shield`'s own per-attempt timeout plus standard Python timeouts.

## Profile defaults

| Field | `internal` | `api` (default) | `slow` |
|---|---|---|---|
| `max_consecutive_failures` | 10 | 20 | 5 |
| `initial_max_rate` | 100/s | 2/s | 0.5/s |
| `adaptive_burst_capacity` | 200 | 5 | 1 |
| `min_rate_floor` | 1/s | 0.25/s | 0.05/s |
| `initial_timeout` | 1.0s | 10.0s | 120.0s |
| `timeout_clamp_min` | 0.05s | 0.5s | 5.0s |
| `timeout_clamp_max` | 5.0s | 60.0s | 600.0s |
| `backoff_scale` | 0.5 | 1.0 | 2.0 |
| `backoff_cap` | 5s | 30s | 60s |

Algorithm invariants (do not vary by profile): `max_attempts = 4`, CUBIC `C = 0.4`, `β = 0.7`, measured-rate clamp `1.5×`, exponential full-jitter backoff, p95 × 2.5 adaptive timeout estimator over a 32-sample window.

## Composing with client-side retries

The profile defaults assume one layer of client-side retries below. `backoff_scale` is sized to operate on a slower timescale than typical inner-layer backoff envelopes. No need to disable the client's retries:

```python
@Shield.api(timeout_errors=(requests.Timeout, requests.ConnectionError))
def call_api(...): ...

@Shield.internal(timeout_errors=(botocore.exceptions.ReadTimeoutError,
                                    botocore.exceptions.ConnectTimeoutError,
                                    botocore.exceptions.EndpointConnectionError))
def get_object(...): ...

@Shield.slow(timeout_errors=(netmiko.exceptions.NetmikoTimeoutException,
                                    netmiko.exceptions.ReadTimeout))
def run_command(...): ...
```

## File layout

- `grelmicro/resilience/shield.py` — public `Shield` class, `shield` module-level decorator.
- `grelmicro/resilience/_retry_budget.py` — `asyncio.Lock`-protected consecutive-failure counter.
- `grelmicro/resilience/_adaptive_gate.py` — lazy-enable token bucket, CUBIC controller, rate-clocker.
- `grelmicro/resilience/_timeout_estimator.py` — 32-sample rolling p95 with clamp.
- `grelmicro/resilience/_shield_profile.py` — `ShieldProfile` pydantic model and the three built-in instances.
- `grelmicro/resilience/__init__.py` — export `Shield`, `ShieldProfile`, `shield`.

## Out of scope

- **Hedged / tied requests** (Dean & Barroso 2013). Deserves its own primitive.
- **Distributed retry budget** across processes. Per-instance, per-process only in v1.
- **Deadline propagation** into the wrapped callable. Per-attempt timeouts only. Users wrap the whole call in `asyncio.timeout(...)` if they need a total deadline.
- **HTTP status sniffing / `Retry-After` header parsing.** Spec is exception-only. Users that have a response envelope `raise_for_status()` and pass the resulting exception type via `timeout_errors=`.

## Acceptance

- [ ] `Shield` class with `internal` / `api` / `slow` factory classmethods, `from_config`, `run`, and `__call__` (decorator form).
- [ ] Module-level `shield` decorator with `.internal()`, `.api()`, `.slow()` variants.
- [ ] `ShieldProfile` pydantic model with all nine fields and the three built-in instances.
- [ ] Exception classification matches the API contract table exactly: only `timeout_errors` subclasses are retried; everything else propagates.
- [ ] Retry-budget bucket: `asyncio.Lock`-protected counter, capacity = `profile.max_consecutive_failures`, `+1` refund per recovered retry, denied acquire returns False (never raises), PEP 678 note on surfaced exceptions when the budget caused the stop.
- [ ] Adaptive bucket: disabled until first `timeout_errors` exception; once enabled, blocking acquire on every call; CUBIC controller with `C = 0.4`, `β = 0.7` from RFC 9438; rate-clocker for measured-rate clamp at `1.5×`.
- [ ] Per-attempt timeout: p95 of last 32 successful latencies × 2.5, clamped to profile range, initial value from profile.
- [ ] Backoff: exponential full jitter `random_uniform(0, 1) · min(profile.backoff_scale · 2^(i−1), profile.backoff_cap)`.
- [ ] `BaseException` outside `Exception` always propagates without retry; `ResilienceError` subclasses always propagate.
- [ ] Tests: classification table validated exception-by-exception; refund rules; lazy-enable on first slow-down; CUBIC numbers pinned to formulas from the spec; cancellation propagates through every layer.
- [ ] `docs/resilience/shield.md` with the four-line example, the classification table, and pointers to the architecture spec.
- [ ] `docs/resilience/index.md` updated to list `Shield` as the recommended default.
- [ ] `docs/changelog.md` entry.

## References

- Spec: `docs/architecture/shield-spec.md`
- RFC 9438 (CUBIC): https://datatracker.ietf.org/doc/rfc9438/
- Bronson et al., \"Metastable Failures in Distributed Systems\" (HotOS 2021): https://dl.acm.org/doi/10.1145/3458336.3465286
- Huang et al., \"Metastable Failures in the Wild\" (OSDI 2022): https://www.usenix.org/conference/osdi22/presentation/huang-lexiang
- PEP 654 (exception groups), PEP 678 (exception notes), PEP 789 (yield inside cancel scopes).

grelinfo:
## Summary

Closes #249. Adds `Shield`, the outer-layer resilience primitive for one async call to one dependency. Bundles per-attempt adaptive timeout, exponential-jittered retry gated by a consecutive-failure budget, and a CUBIC-style adaptive rate limiter that engages only after sustained slow-downs. One decorator, one knob (`timeout_errors=`).

```python
@shield.api(timeout_errors=(httpx.TimeoutException, httpx.ConnectError))
async def fetch(url: str) -> bytes: ...
```

Three profiles ship (`internal`, `api`, `slow`). Algorithm parameters are profile-fixed (CUBIC C=0.4, β=0.7; 4 attempts; p95×2.5 over the last 32 successes; exponential full jitter). The only user knobs are `timeout_errors`, `max_rate`, `cache`, `cache_key`, `fallback`, `name`.

Cache + fallback chain on give-up: cache hit → cached value, cache miss → fallback callable, neither → raise with a PEP 678 note.

## File layout

```
grelmicro/resilience/shield/
├── __init__.py
├── _profile.py            # _BaseShieldConfig (public fields)
├── _internal.py / _api.py / _slow.py  # discriminated profile configs
├── _retry_budget.py       # asyncio.Lock-protected counter + refunds
├── _adaptive_gate.py      # token bucket + CUBIC controller + measured-rate clocker
├── _timeout_estimator.py  # 32-slot p95×2.5 ring buffer
├── _key.py                # stable_hash + default_cache_key
├── _shield.py             # Shield class
└── _decorator.py          # module-level @shield with .internal/.api/.slow
```

## Spec ambiguities resolved

1. Cache miss = `cache.get(key)` returns `None`. Distinguished from "cached None" by sentinel inside the helper.
2. Adaptive bucket zeros tokens on first slow-down so the new ceiling engages immediately.
3. Measured-rate clamp returns `inf` (disengaged) with fewer than 2 samples.
4. Refund-on-recovery refunds the total retries actually consumed on that call.
5. `Shield.run(sync_fn)` raises `TypeError` with a hint about `asyncio.to_thread`.
6. CUBIC shrink fires per failed attempt (matches the spec wording).
7. Reconfigure builds fresh helpers; in-flight calls keep their captured snapshot.

## Scope

In scope:
- Three profiles, factory classmethods, `Shield.from_config`, `Shield.run`, `Shield.__call__`.
- Module-level `shield` decorator (`@shield`, `@shield.internal/api/slow`).
- Cache + fallback chain.
- Live reconfigure.
- Env loading via `GREL_SHIELD_{NAME_UPPER}_*`.
- Snippets convert all inline code blocks in `shield.md`.

Out of scope (per spec): hedged requests, distributed retry budget, deadline propagation, HTTP status sniffing, sync support.

## Test plan

- [x] 95 new tests under `tests/resilience/shield/` covering profile defaults, retry budget refunds, CUBIC numbers pinned to formulas, p95 estimator window, classification table, cache fallback chain, decorator forms, env loading.
- [x] `pytest tests/resilience/shield/` — 95 passed, 100% coverage on the new package.
- [x] `pytest` (full suite) — 1869 passed.
- [x] `ruff check .` clean.
- [x] `ty check` clean for Shield code.
- [x] `mkdocs build --strict` passes.
- [ ] First Copilot review pass.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-* ✨ Add `Shield` resilience pattern: per-attempt timeout, retry-budget-gated retries, CUBIC-style adaptive rate limiter, optional cache and fallback recovery paths. Three profiles (`internal`, `api`, `slow`) cover the common cases. Decorator (`@shield`, `@shield.api(...)`), class (`Shield.api("name")`), and imperative (`shield.run(fn, ...)`) forms supported. Issue [#249](https://github.com/grelinfo/grelmicro/issues/249).
+* ✨ Add `Shield` resilience pattern: per-attempt timeout, retry-budget-gated retries, CUBIC-style adaptive rate limiter, optional cache and fallback recovery paths. Three profiles (`internal`, `api`, `slow`) cover the common cases. Decorator (`@shield`, `@shield.api(...)`), class (`Shield.api("name")`), and imperative (`Shield.api("name").run(fn, ...)`) forms supported. Issue [#249](https://github.com/grelinfo/grelmicro/issues/249).
 
 ## 0.25.0 - 2026-05-21
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+* ✨ Add `Shield` resilience pattern: per-attempt timeout, retry-budget-gated retries, CUBIC-style adaptive rate limiter, optional cache and fallback recovery paths. Three profiles (`internal`, `api`, `slow`) cover the common cases. Decorator (`@shield`, `@shield.api(...)`), class (`Shield.api("name")`), and imperative (`shield.run(fn, ...)`) forms supported. Issue [#249](https://github.com/grelinfo/grelmicro/issues/249).
+
 ## 0.25.0 - 2026-05-21
 
 ### Features
--- a/docs/reference/resilience.md
+++ b/docs/reference/resilience.md
@@ -39,6 +39,8 @@
         - RateLimitResult
         - RedisCircuitBreakerAdapter
         - RedisRateLimiterAdapter
+        - ApiShieldConfig
+        - InternalShieldConfig
         - ResilienceError
         - ResilienceSettingsValidationError
         - Retry
@@ -46,7 +48,10 @@
         - RetryBackoffConfig
         - RetryConfig
         - RetryStrategy
+        - Shield
+        - ShieldConfig
         - SlidingWindowConfig
+        - SlowShieldConfig
         - Timeout
         - TimeoutConfig
         - TokenBucketConfig
@@ -54,3 +59,4 @@
         - falling_back
         - retry
         - retrying
+        - shield
--- a/docs/resilience/index.md
+++ b/docs/resilience/index.md
@@ -2,6 +2,7 @@
 
 The `resilience` package provides primitives that help your services handle failures in distributed systems. Each pattern is independent. Use the ones you need and skip the rest.
 
+- [**Shield**](shield.md): all-in-one outer-layer resilience for one async call to one dependency. Bundles per-attempt timeout, exponential-jittered retry with a consecutive-failure budget, and CUBIC-style adaptive throttling. One decorator, one knob. The recommended starting point.
 - [**Circuit Breaker**](circuit-breaker.md): detect repeated failures in a downstream call and stop sending requests long enough for it to recover. This prevents cascading failures.
 - [**Fallback**](fallback.md): return a safe default value when a call fails. Pluggable filter (any `Match`), static default or factory callable, decorator and block forms.
 - [**Rate Limiter**](rate-limiter.md): cap how many requests a client can make per window. Pluggable algorithm (`TokenBucketConfig` or `SlidingWindowConfig`), pluggable backend (`Memory` or `Redis`), with a result shape that maps directly to HTTP rate limit headers.
@@ -14,6 +15,7 @@ See [Composing patterns](composition.md) for the recommended outside-in order wh
 
 | Pattern | Use when |
 |---|---|
+| **Shield** | You wrap one async call to one dependency and want sensible-by-default resilience without composing primitives. |
 | **Circuit Breaker** | The caller is hitting an external dependency (DB, API, third-party) that can fail or stall. |
 | **Fallback** | A call can fail and you want to degrade gracefully (cached value, empty list, neutral default) instead of propagating the error. |
 | **Rate Limiter** | You need to throttle *your own* endpoint, worker, or background job to protect the downstream side or enforce fair usage. |
--- a/docs/resilience/shield.md
+++ b/docs/resilience/shield.md
@@ -1,0 +1,295 @@
+# Shield
+
+A **Shield** is the outer layer of resilience around one async call to one dependency. It bundles a per-attempt timeout, exponential-jittered retry gated by a consecutive-failure budget, and a CUBIC-style adaptive rate limiter that engages only after sustained slow-downs. One decorator, one knob.
+
+**Why**
+
+- Wrap one call to one dependency without composing four primitives by hand.
+- Survive transient slow-downs (network blips, brief overloads) without changing the call site.
+- Pace the whole client down automatically when a dependency keeps timing out, then ramp back gradually.
+- Stay invisible on the happy path. The adaptive layer enables itself only after the first slow-down.
+
+The adaptive layer uses a CUBIC-style controller from [RFC 9438](https://datatracker.ietf.org/doc/rfc9438/) with `C = 0.4` and `β = 0.7`. The per-attempt timeout tracks the p95 of the last 32 successful latencies multiplied by `2.5`, clamped to the profile range. Retries are gated by a consecutive-failure budget with refund-on-recovery.
+
+## Usage
+
+```python
+import httpx
+from grelmicro.resilience import shield
+
+
+@shield.api(timeout_errors=(httpx.TimeoutException, httpx.ConnectError))
+async def fetch(client: httpx.AsyncClient, url: str) -> bytes:
+    response = await client.get(url)
+    response.raise_for_status()
+    return response.content
+```
+
+Pick a profile factory (`shield.internal`, `shield.api`, `shield.slow`), tell Shield which exceptions mean "the dependency is slow", done. Everything else propagates unchanged.
+
+The zero-argument form uses the `api` profile with the default `timeout_errors=(TimeoutError,)`:
+
+```python
+@shield
+async def cheap_call() -> None:
+    async with asyncio.timeout(5):
+        await do_work()
+```
+
+Shield is **async-only**. Use it on coroutine functions.
+
+## Choose a profile
+
+| Profile | Use when | Initial rate (per second) | Initial timeout |
+|---|---|---|---|
+| `internal` | In-cluster RPC. Healthy services, fast latency, tight budgets. | `100` | `1.0s` |
+| `api` (default) | External HTTP APIs. Moderate latency, occasional outages, third-party SLAs. | `2` | `10.0s` |
+| `slow` | Long-running calls: LLM inference, batch jobs, large queries. | `0.5` | `120.0s` |
+
+Three profiles cover the common cases. They are mutually exclusive: pick one per `Shield` instance. To pace differently, switch profile, do not tune individual fields. The internal parameters (retry budget capacity, CUBIC curve constants, backoff scale, timeout clamps) are fixed by profile choice.
+
+```python
+@shield.internal(timeout_errors=(MyRpcTimeout,))
+async def call_internal_rpc(): ...
+
+@shield.api(timeout_errors=(httpx.TimeoutException, httpx.ConnectError))
+async def call_external_api(): ...
+
+@shield.slow(timeout_errors=(MyLLMError,))
+async def call_llm(prompt: str): ...
+```
+
+If the three profiles do not fit, you have outgrown Shield. Compose `Retry`, `Timeout`, `RateLimiter` and `CircuitBreaker` directly instead.
+
+## Exception classification
+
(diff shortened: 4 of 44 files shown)


grelinfo:
Thanks @copilot. Verdicts on the 6 comments, fixed where valid in 77d6e6d:

**1. Helpers swapped mid-call by reconfigure** — Valid bug. Fixed by moving `_retry_budget`, `_adaptive_gate`, `_timeout_estimator` into the frozen `_State` dataclass. `_execute` captures one `state` reference at the top and reads helpers from it end-to-end. `_handle_give_up`, `_compute_key`, `_fire_and_forget_cache_set`, and `_backoff_for` now take `state` as a parameter. `_apply_reconfigure` builds a fresh `_State` and swaps a single reference. In-flight calls keep their snapshot.

**2. Cache `None` vs miss conflation** — Won't fix. `grelmicro.cache.TTLCache.get(key, default=None)` returns the default on miss, which is exactly Python's `dict.get` convention and matches every Python cache library (`functools.lru_cache`, `cachetools`, `aiocache`, `diskcache`). Cached `None` being indistinguishable from miss is the conventional contract. Documenting it in the cache section of `shield.md` would be the right cleanup if it surprises users in practice, but the implementation matches the convention.

**3. Strong reference / GC race on cache-write task** — Valid. Fixed by storing each cache-write task in a new `self._pending_tasks: set[asyncio.Task]`, with `task.add_done_callback(self._pending_tasks.discard)`. This is the asyncio docs' recommended pattern for fire-and-forget tasks.

**4. `_discard_task` swallows only `Exception`, not `BaseException`** — Valid. Fixed by dropping `_discard_task` entirely. The new `_pending_tasks.discard` callback never raises and never reads `task.result()`, so cancellation no longer needs explicit suppression. `_safe_cache_set` continues to absorb every `Exception` from the cache backend.

**5. Eager `from .shield import shield` breaks the lazy-load promise in the package docstring** — Valid (for `retry` / `fallback` too — pre-existing). Fixed the docstring at `grelmicro/resilience/__init__.py` to list the eager exports explicitly (`retry`, `retrying`, `fallback`, `falling_back`, `shield`) and acknowledge the subpackage load for `shield`. The architectural fix (lazy decorator with deferred Shield import) is out of scope for this PR.

**6. `_resolve_fqn` accepts `BaseException` but the message says `Exception`** — Valid. Tightened the validation to `issubclass(cls, Exception)` so the check now matches the message and the classification contract (`BaseException` outside `Exception` always propagates, so accepting it in `timeout_errors` was meaningless).

Test impact: 95 shield tests + 1869 full suite still green; ruff, ty, mkdocs strict all clean.

grelinfo:
Round 2 review addressed in 6263aa3:

**1. `ResilienceError` swallowed by cache/fallback / mutated with PEP 678 note** — Valid bug. Fixed: `ResilienceError` now `raise`s immediately from `_execute`, mirroring the `BaseException`-outside-`Exception` treatment. No cache lookup, no fallback call, no note. Added `test_resilience_error_skips_cache_and_fallback` that constructs a Shield with both `cache=` and `fallback=` set, raises a `ResilienceError`, and asserts the original error propagates with no `__notes__` attached.

**2. None vs miss (recurrence)** — Won't fix, same reply as round 1. `grelmicro.cache.TTLCache.get(key, default=None)` already conflates cached `None` with miss, matching every Python cache library.

**3. `[PEP 654] groups` wording in shield.md** — Valid. Replaced with `BaseExceptionGroup` (the only group type outside `Exception`), dropped the PEP 654 link from that cell.

**4. `shield_giveup.py` never raises** — Valid. Snippet now wraps `fetch` with `@shield.api(timeout_errors=(httpx.TimeoutException,))` and the function body raises `httpx.TimeoutException` so the example output is actually reachable. Added `EM101` / `TRY` ignores under `[tool.ruff.lint.per-file-ignores]` for that snippet, matching the pattern already used for `docs/snippets/logging/basic.py`.

**5. Changelog mentions non-existent `shield.run`** — Valid. Reworded to `Shield.api("name").run(fn, ...)` to point at the actual imperative form on the class.

Status: 96 shield tests + 1870 full suite green, ruff / ty / mkdocs strict clean.

grelinfo:
Round 3 review addressed in 3d0622b:

**1. None vs miss (3rd recurrence)** — Won't fix. `grelmicro.cache.TTLCache.get(key, default=None)` returns the default on miss; the conventional Python cache contract is "cached `None`" indistinguishable from "miss". Same as `dict.get`, `functools.lru_cache`, `cachetools`, `aiocache`. Documenting this in shield.md would be the right cleanup if it ever surprises users in practice, but the implementation matches the convention every Python cache library uses.

**2. Non-retryable Exception goes through give-up path (cache/fallback, PEP 678 note)** — Intentional, but the classification table was ambiguous about it. **Fixed the doc:** the table now has a fifth column ("Cache / fallback recovery?") that explicitly says yes for non-retryable Exception, no for `ResilienceError` and `BaseException` outside `Exception`. The prose under the table now reads: "Any other `Exception` skips the retry loop and goes straight to the give-up recovery path: if a cache or fallback returns a value, that value is returned; otherwise the original exception is re-raised with a PEP 678 note. `ResilienceError` and `BaseException`-outside-`Exception` always propagate unchanged with no recovery and no note." The implementation matches.

**3. `_safe_cache_set` docstring says "every exception" but only catches `Exception`** — Valid. Tightened the docstring: "Swallow every `Exception` at debug. `BaseException` (notably `asyncio.CancelledError`) propagates so task cancellation during shutdown reaches the asyncio scheduler unchanged."

**4. `_normalize_timeout_errors` accepts `BaseException` for class-direct input** — Valid. Tightened to `issubclass(value, Exception)` with a clear error: `"timeout_errors entry <class 'KeyboardInterrupt'> is not an Exception subclass; BaseException-only types are never retried."` Added `test_timeout_errors_rejects_base_exception_class` covering the new path.

97 shield tests + 1870 full suite green; ruff / ty / mkdocs strict clean.

PR #290, opened 2026-05-23
✨ R1 Public API: TTLCacheConfig, minimal example, fixes

grelinfo:
## Summary

Addresses Finding #1-#5 from the R1 Public API Designer road-to-1.0 review.

- ✨ Add `TTLCacheConfig` (frozen Pydantic) and expose it via `TTLCache.config`, matching every other primitive (Finding #5).
- 📝 Add a one-route, one-primitive FastAPI example above the full composition demo in `README.md` and `docs/index.md` (Finding #2).
- 📝 Replace stale `RateLimit` import/usage with `RateLimiters` in both `README.md` and `docs/index.md` (Finding #1).
- 📝 Annotate `Grelmicro.use`, `Grelmicro.get`, `instrument`, and `CacheBackend.{get,set,delete}` parameters with `Annotated[..., Doc(...)]` (Finding #3).
- 📝 Update the `CONTRIBUTING.md` discriminator rule from `type` to `kind` to match the deliberate 0.23.0 rename ([#268](https://github.com/grelinfo/grelmicro/issues/268)) that freed the Python `type` builtin (Finding #4 reversed: code was right, rule was stale).

## Test plan

- [x] `pytest` (1872 passed, +1 new `TTLCacheConfig` test)
- [x] `mkdocs build --strict`
- [x] `ty check` clean on touched modules
- [x] Verified `from grelmicro.resilience import RateLimiters` imports cleanly
- [x] Verified minimal README example imports and constructs without a `Grelmicro(...)` app

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,9 +110,10 @@ The rules below are the ones the tooling can't check:
 - Every configuration class is
   `BaseModel, frozen=True, extra="forbid"`.
 - Every field is annotated with `Annotated[type, Doc("...")]`.
-- For discriminated unions, include a `type: Literal["..."]`
+- For discriminated unions, include a `kind: Literal["..."]`
   discriminator field and wrap the union with
-  `Annotated[A | B, Discriminator("type")]`.
+  `Annotated[A | B, Discriminator("kind")]`. `kind` avoids
+  shadowing the Python `type` builtin on every config object.
 - Expose the config through a `@property def config(self) -> XConfig`
   on the front-door class.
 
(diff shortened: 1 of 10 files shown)


grelinfo:
Thanks for the catch — verified `acquire` requires `key=` and returns `RateLimitResult` (the `RateLimitExceededError` raise path lives on `acquire_or_raise`). Pushed [6ada1b2](https://github.com/grelinfo/grelmicro/commit/6ada1b2) which switches the minimal example to `await api_limiter.acquire_or_raise(key=request.client.host)` in both `README.md` and `docs/index.md`.

grelinfo:
Addressed the follow-up comments in [118ffb9](https://github.com/grelinfo/grelmicro/commit/118ffb9):

- Switched the minimal example to `key="global"` per the original suggestion. The constant key sidesteps `request.client` being `None` and keeps the example as the smallest possible demo.
- Updated `TTLCache` class docstring to call out `pydantic.ValidationError` (still a `ValueError` subclass).
- Updated the three TTLCache invalid-input tests to assert `ValidationError` so they fail loudly if the validation mechanism drifts.

PR #291, opened 2026-05-23
🔧 Cleanup batch: ty, docs anchor, trace timeout, secret redaction

grelinfo:
## Summary

Low-risk wins from the road-to-1.0 reviews bundled into one PR.

### Static & docs (R2, R4, R8)
- 🔧 Drop three unused `ty: ignore` directives in `grelmicro/_json.py`.
- 📝 Fix the broken `sync.md → task.md#tasks` anchor flagged by `mkdocs --strict`.
- ✅ Guard test: every `_LAZY` key in `grelmicro/resilience/__init__.py` is in `__all__` and resolves at runtime.

### Reliability (R7)
- ✨ Add `TracingConfig.shutdown_timeout` (default `5.0s`). `Trace.__aexit__` runs `TracerProvider.shutdown()` in a thread bounded by that deadline. A slow or broken OTLP exporter no longer hangs application shutdown.

### Security (R9)
- 🔒 `SettingsValidationError` no longer echoes the offending input value. Env-loaded credentials (DSNs, tokens) stop leaking into error messages.
- ✨ Add `RedisProvider.safe_url` and `PostgresProvider.safe_url`. The new `__repr__` on both providers uses the safe form so passwords never leak through logs or tracebacks.

### Docs (R3)
- 📝 Document `Tasks` as per-process and point at `TaskLock` / `LeaderElection` for cluster-wide scheduling.
- 📝 Add a Kubernetes operational-assumptions section to `docs/architecture/kubernetes.md`: RBAC, API server availability, etcd latency, single-cluster scope.

## Test plan

- [x] `pytest` (1884 passed, +12 new tests)
- [x] `mkdocs build --strict` clean
- [x] `ty check` reports `All checks passed!`
- [x] Manual verification: `RedisProvider(URL_WITH_PASSWORD).safe_url` redacts; `repr()` never includes the password
- [x] Manual verification: simulated slow OTel shutdown emits the expected timeout warning and never blocks `async with micro:`

### DIFF
--- a/docs/architecture/kubernetes.md
+++ b/docs/architecture/kubernetes.md
@@ -52,3 +52,28 @@ This prepends `myapp-` to every lease name before sanitization, ensuring differe
 ## Client Library
 
 The backend uses [lightkube](https://lightkube.readthedocs.io/) as the Kubernetes async client. lightkube is lightweight, async-native (built on httpx), and provides strong typing for Kubernetes resources.
+
+## Operational Assumptions
+
+The Kubernetes backend depends on the cluster control plane for every primitive call. Plan deployments around these assumptions:
+
+- **Single cluster**: Locks coordinate within one Kubernetes cluster. Multi-cluster fleets need a backend that fronts a shared store (Redis, Postgres).
+- **API server availability**: Every `acquire`, `extend`, `release`, `locked`, and `owned` call hits the kube-apiserver. If the control plane is unreachable, the operation fails. Acquire-time errors surface as `LockAcquireError`. Pair the backend with retries or fall back to a Memory adapter for self-healing flows.
+- **etcd latency**: Lease writes are replicated through etcd, so per-call latency is at least one Raft round-trip. Expect tens to low hundreds of milliseconds, not microseconds.
+- **RBAC**: The Pod's ServiceAccount must hold a Role granting `get`, `create`, `update`, `delete`, and `list` on `leases.coordination.k8s.io` in the target namespace. A minimal Role is:
+
+    ```yaml
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: Role
+    metadata:
+      name: grelmicro-leases
+      namespace: default
+    rules:
+      - apiGroups: ["coordination.k8s.io"]
+        resources: ["leases"]
+        verbs: ["get", "list", "create", "update", "delete"]
+    ```
+
+    Bind it to the workload's ServiceAccount with a `RoleBinding` in the same namespace.
+
+- **Clock skew**: Lease expiry uses the API server's clock via `renewTime + leaseDurationSeconds`. Large Node clock skew (>1 second) can produce surprising acquire/release decisions on Pods that compare against their local clock.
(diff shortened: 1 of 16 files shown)


grelinfo:
Addressed all three follow-up comments in [6da90dc](https://github.com/grelinfo/grelmicro/commit/6da90dc):

1. **Daemon-thread shutdown** (`_component.py`): replaced `asyncio.to_thread(shutdown)` with a `threading.Thread(..., daemon=True)` + `Event.wait(timeout)`. The wait runs in the loop's default executor, which always returns within the timeout. An abandoned slow shutdown thread can no longer block loop teardown or process exit. Test simplified back to `time.sleep(0.3)` with `shutdown_timeout=0.05` (proving the daemon-thread path).
2. **Log precision**: switched the timeout warning from `%.1fs` to `%ss` so `0.05`-style timeouts log faithfully.
3. **Regex fallback in `_redact_url`** (both providers): on `Url(url)` / `MultiHostUrl(url)` parse failure, a conservative `://[^/?#@]*:([^@/?#]+)@` regex replaces the password with `***` before returning. A malformed DSN that still contains credentials no longer leaks via `safe_url` / `repr()`. New tests cover the malformed-with-password path for both providers.

Coverage: 1886 tests pass, `ty check` clean, mkdocs strict still green.

grelinfo:
Addressed all 3 follow-up comments in [aae5d95](https://github.com/grelinfo/grelmicro/commit/aae5d95):

1. **Postgres multi-host redaction**: regex updated from `(://...)` to `(://|,)(...)@` so every `user:pass@` pair in a malformed multi-host DSN is redacted, not just the first. New test `test_safe_url_malformed_multi_host_redacts_every_password` proves it.
2. **`shutdown_timeout` docstring**: rewritten to match actual behavior — "On timeout the call is abandoned (the daemon shutdown thread keeps running but cannot block loop teardown), a warning is logged, and the rest of `__aexit__` proceeds. Pending spans may be dropped."
3. **`caplog` set_level**: both trace-shutdown tests now call `caplog.set_level("WARNING", logger="grelmicro.trace._component")` so the assertions don't depend on global log config.

grelinfo:
Addressed all 4 follow-up comments in [00536ce](https://github.com/grelinfo/grelmicro/commit/00536ce):

1. **Log-forging safe repr**: `__repr__` on both providers now uses `{self.safe_url!r}` so quotes/newlines/non-printables in a URL can't produce misleading or forged output.
2. **Narrow exception catch**: `_run_with_timeout._runner` now catches `Exception` (not `BaseException`) so `KeyboardInterrupt` and `SystemExit` aren't suppressed inside the shutdown thread.
3. **Tab heading**: `docs/sync.md` "Task Manager (Recommended)" → "Tasks (Recommended)" for consistency with the prose and `docs/task.md`.
4. **Generic error prefix**: `SettingsValidationError` message is now "Could not validate settings:" (was "...environment variables settings:") since the class is used for kwargs validation too. Test matchers updated.

All 1888 tests pass.

grelinfo:
Rebased onto main after #290 merged (changelog conflict resolved). Copilot review #5 on the prior tip (00536ce) was clean. Tests/ty/mkdocs all still pass on the rebased branch.

grelinfo:
Addressed both Copilot follow-ups in [ebc2663](https://github.com/grelinfo/grelmicro/commit/ebc2663):

- Added `_redact_query()` to both providers that case-insensitively redacts common credential-like query keys: `password, passwd, pwd, token, access_token, auth, secret, client_secret, api_key, apikey, key`.
- `_redact_url()` now consults this in addition to the userinfo redaction, so URLs like `redis://host/0?password=secret` and `postgresql://host/db?token=abc` no longer leak.
- New tests cover redaction, case-insensitivity, and the pass-through path (query without credentials).

1893 tests pass, ty clean.

PR #359, opened 2026-06-12
✨ Harden the API for the 1.0 RC: ExternalConfig, middleware, one-rule resolution, Idempotency, Valkey, Sentinel/Cluster

### LINKED ISSUE #154: 💭 Externalised configuration integration (ConfigSource, bind_source)
## Context

The [Externalised Configuration pattern](https://microservices.io/patterns/externalized-configuration.html) describes services that read and reload their settings from an external store (Kubernetes `ConfigMap`, Consul, etcd, AWS Parameter Store, Spring Cloud Config Server, ...).

`Component.reconfigure(new_config)` (#120) gives us the live-update primitive. This issue tracks the optional, additive surfaces that would make plugging an external config store into grelmicro a one-liner. **Unscheduled.** Open for discussion, not committed to any milestone.

## What is already in place

- `await component.reconfigure(new_config)` on every reconfigurable primitive.
- Atomic snapshot publish, in-flight isolation, same-type guard, equality short-circuit (see [Live reconfiguration](https://grelinfo.github.io/grelmicro/architecture/reconfigure.html)).
- Per-instance identity via `name`, frozen Pydantic configs that round-trip through JSON.

## What this issue is about

Four additive surfaces. Each is independently shippable, each is non-breaking, none of them changes the existing `reconfigure` contract. Listed as ideas, not commitments.

### 1. `_on_reconfigured(old, new)` hook

Default no-op on the `Reconfigurable` mixin. Lets observability layers (audit log, metrics, tracing) react to swaps without monkey-patching.

```python
class AuditedRateLimiter(RateLimiter):
    async def _on_reconfigured(self, old, new):
        audit_log.info("rate limiter %s reconfigured", self.name, extra={"old": old, "new": new})
```

### 2. `ConfigSource[ConfigT]` Protocol

The contract an external config store implements.

```python
from typing import Protocol, AsyncIterator
from pydantic import BaseModel

class ConfigSource(Protocol[ConfigT]):
    async def watch(self, name: str) -> AsyncIterator[ConfigT]:
        """Yield a fresh config every time the external store reports a change."""
```

### 3. `bind_source(component, source)` helper

Reads from the source and forwards each yielded config to `component.reconfigure(...)`. Catches and logs validation errors so a bad config in the external store does not kill the watcher.

```python
async def main():
    limiter = RateLimiter.token_bucket("api", capacity=10, refill_rate=1)
    source = ConfigMapSource(namespace="prod", name="rate-limits")  # user-provided
    async with anyio.create_task_group() as tg:
        tg.start_soon(bind_source, limiter, source)
        # ... rest of the app
```

### 4. `ReconfigurableRegistry`

Optional opt-in. Weak-ref dict keyed by `(type, name)`. Lets an external orchestrator iterate every reconfigurable component without the application maintaining its own map.

```python
for component in ReconfigurableRegistry.iter():
    await component.reconfigure(load_from_external_store(component.name))
```

## Out of scope here

- Shipping a Kubernetes `ConfigMap` or Consul or etcd watcher inside the library. Those bring heavy dependencies and are best left to per-store companion packages.
- Hot-swapping the backend from the new config (already out of scope per the architecture doc).

## Acceptance for "we are ready"

When all four surfaces (or a documented subset) are in place, integrating an external config store should be:

1. Implement `ConfigSource[ConfigT]` against the store.
2. Call `bind_source(component, source)` for each component.
3. Done. No grelmicro internals touched.

## References

- [microservices.io: Externalized Configuration](https://microservices.io/patterns/externalized-configuration.html)
- #120 (`Component.reconfigure`, foundation for this work)
- [Live reconfiguration architecture](https://grelinfo.github.io/grelmicro/architecture/reconfigure.html)

### LINKED ISSUE #328: ✨ Make Grelmicro.current() ambient resolution work in FastAPI request handlers
## Problem

`Grelmicro.current()` reads a per-asyncio-task ContextVar set in the lifespan task. FastAPI/Starlette request handlers run in **separate tasks** that do not inherit it, so patterns resolving their backend ambiently inside a request handler fail:

- `Lock` / `TTLCache` raise `NoActiveAppError`.
- `RateLimiter` / `CircuitBreaker` silently fall back to an in-process memory adapter (wrong backend, no error).

Background `Tasks` are unaffected (they are children of the lifespan task and inherit the ContextVar).

This was found while building the FastAPI demo (#166). As a minimal fix, the README and `simple_fastapi_app.py` examples now pass an explicit `backend=` to patterns used in handlers, and the demo does the same. But ambient resolution in handlers is the more ergonomic story and currently does not work.

## Options

1. **FastAPI dependency**: a `Depends` that enters/sets the active app per request.
2. **ASGI middleware**: set `_current_micro` for the duration of each request.
3. **Document explicit backends as the FastAPI pattern** and drop the ambient promise for handlers.

## Acceptance

- Decide the approach (dependency / middleware / explicit-only).
- If ambient is supported, a request handler can use `Lock("x")` / `@cached(cache)` without an explicit backend and resolve the registered backend.
- Update the README front-door example and `simple_fastapi_app.py` accordingly.

Deferred from the 0.26.0 release; the examples are corrected with explicit backends in the meantime.

grelinfo:
## What this ships

One PR carrying the RC hardening pass from the API review (`API_REVIEW_RC.md`).

### New patterns and modules
- **ExternalConfig** (`grelmicro.config`): live reconfiguration from a mounted ConfigMap, Secret, `.env`, JSON, YAML, or TOML source. Every named pattern is addressable by its `GREL_{PATTERN}_{NAME}_` keys, including `CircuitBreaker` and `RateLimiter`. Public `reload()` for deterministic passes. Validation warnings log field names only, never values.
- **Idempotency** (`grelmicro.idempotency`): caller-keyed execute-once with replay, single-flight duplicates across replicas, nothing stored on failure, optional payload fingerprint. Rides the cache layer.
- **GrelmicroMiddleware** (`grelmicro.fastapi`): pure ASGI, binds the active app inside request handlers so ambient resolution works there.

### Breaking (pre-1.0 clean cuts)
- One backend resolution rule everywhere: explicit `backend=`, else the active app's component, else `OutOfContextError`. The implicit memory fallback on `CircuitBreaker` and `RateLimiter` is gone.
- `testing.Call` is now `Call(method, args=..., kwargs=...)`.

### Coordination and scheduler
- `Lock.acquire(timeout=)`, `Lock.extend()`, `TaskLock.refresh()`, `retry_jitter` on `Lock` and `LeaderElection`.
- Scheduler introspection: `next_fire_time` and `last_fire` (`FireInfo`) on interval and cron tasks.

### Backends
- **Valkey**: `ValkeyProvider` (extra `valkey`) serves the full Redis adapter column.
- **Redis Sentinel and Cluster**: `redis+sentinel://` and `redis+cluster://` URL schemes plus `RedisProvider.sentinel(...)` / `.cluster(...)` factories. Multi-key cache and lock operations on Cluster require a hash-tagged prefix, enforced at construction.

### Polish
- Fencing token guarantees documented per backend with a worked resource-side check.
- `Match.explain()`, non-bool predicate warning, positional args in `testing.record`, actionable sync `@cached` error.
- README, demo, and snippets rewritten to the middleware idiom. New `docs/roadmap.md`.

Combined coverage gate at 100 percent (unit plus integration), ruff, ty, and mkdocs all clean.

Fixes #328
Fixes #174
Fixes #154


grelinfo:
Copilot review triage, all four findings evaluated against the code:

1. **CircuitBreaker registration rule (valid, fixed)**: the registration discriminator was inconsistent across patterns. Unified rule now: a pre-built config object (constructor `config=` or `from_config`) means declarative and stays static, kwargs and factory construction register for live reload. `RateLimiter.__init__` flipped to static, its factories register directly.
2. **`ValkeyProvider.from_client` (valid, fixed)**: now calls `_bind_valkey_classes()` so a Valkey Cluster client is detected and the hash-tag guard applies.
3. **Valkey `:latest` tag (valid, fixed)**: pinned to `valkey/valkey:8.1`.
4. **Bitnami `:latest` tags (valid, fixed)**: Bitnami publishes only `latest` plus digests now, so both images are pinned by manifest digest.

Same commit also fixes two findings from the DX audit: `Idempotency` now raises the documented `OutOfContextError` with an actionable message, and the resilience module docstring no longer advertises a `RateLimiter("name", limit=, window=)` form that does not exist.

PR #360, opened 2026-06-12
✨ Make uses=[...] forgiving: bare classes, provider auto-registration

grelinfo:
## What this ships

Implements the ergonomic `uses=[...]` from #263, on top of the #359 RC hardening.

- **Bare component or adapter class** in `uses=[...]` / `use(...)` is instantiated with no arguments. A bare adapter (class or instance) wraps itself in its matching Component by kind.
- **A Provider listed with no Components** auto-registers one default Component per kind it serves (`Grelmicro(uses=[redis])` is the whole wiring story). Explicit wins, all-or-nothing: any explicit Component makes every Provider lifecycle-only, so the list always reads the way it runs.
- **Two bare Providers with no Components raise `AmbiguousProviderError`** (new export) instead of guessing.
- `ValkeyProvider` auto-registers all four kinds like `RedisProvider`.

README and docs examples simplified accordingly. Combined coverage gate at 100 percent, ruff, ty, mkdocs clean.

Fixes #263


grelinfo:
Copilot review triage, both findings evaluated against the code and valid, fixed in the docs commit:

1. **Bare `Cache` class example**: `Cache` requires a positional source, so the example now uses argument-free classes (`Tasks`, `HealthChecks`) and states the rule (a class with a required argument must be passed as an instance). Verified runnable.
2. **Skip-rule wording**: tightened to the implementation: an unserved kind (factory raises `NotImplementedError`) is skipped, any other factory error propagates so a misconfigured Provider fails loudly.

PR #364, opened 2026-06-14
📝 Alpha soak: docs split, chaos + ASGI tests, simpler examples

grelinfo:
## What this ships

The validation pass on the 1.0.0a1 alpha, all docs/tests/examples (no `grelmicro/` source changes).

### Docs restructured FastAPI-style
- **User Guide** teaches one blessed path: construct with kwargs and factories, `Grelmicro(uses=[redis])`, the FastAPI middleware, `GREL_*` env tuning. New `first-steps.md`, `wiring.md`, `testing.md`.
- **Advanced User Guide** holds everything else: the three-paths config trichotomy, `from_config`, `ExternalConfig`, named components, Sentinel/Cluster, plugins.
- README banner corrected to "1.0 alpha" with `--pre` guidance, install guide gains the prerelease note, README links point at the rendered site.

### Tests against real infrastructure
- `tests/chaos/`: 7 fault-injection tests that kill or pause the real backend mid-traffic and assert the documented behavior (fail-open serves, fail-closed propagates, breaker trips and recovers, cache stale-on-error, Sentinel disruption recovers via Retry).
- `tests/test_asgi_frameworks.py`: proves `GrelmicroMiddleware` on Starlette (HTTP + WebSocket) and Litestar, not just FastAPI.

### Examples simplified
- Every `uses=[redis, X(redis)]` becomes `uses=[X(redis)]`: the provider is auto-discovered from the borrowing component, so it is never listed twice.

### Launch asset
- `docs/img/demo/`: a zero-infra recording app and a vhs tape for the README demo GIF.

Benchmark refresh and the rendered GIF land on this branch shortly (a mutation campaign is holding the machine). Combined coverage gate at 100 percent.


grelinfo:
Copilot review triage, all four valid, all fixed in 6e55aaf:

1. **ASGI test imports private `_current_micro`** (×2): switched to the public `Grelmicro.current()` + `NoActiveAppError`, so the tests assert the supported contract.
2. **Starlette `TestClient` websocket cross-loop fragility**: rewrote the websocket test as a pure-ASGI scope drive on the pytest event loop, so the app and `async with micro:` share one loop (no TestClient thread).
3. **Chaos rate-limiter refill flakiness**: set `refill_rate=0.001` (effectively zero) so the 5-then-throttle bursts stay deterministic even when slow CI advances Redis-side time.

Also in this push: simplified every `uses=[...]` example to drop the redundant bare provider (the component auto-discovers it), per maintainer request.

PR #365, opened 2026-06-15
✅ Add killing tests for coordination jitter, graceful-stop, and error-name gaps

grelinfo:
## What this ships

A mutation campaign (mutmut) on `grelmicro/coordination/{lock,leaderelection,memory}.py` found the suite was assertion-strong on core invariants but line-strong on peripheral paths: 60 mutants survived because tests executed lines without asserting outcomes. This closes the three highest-value clusters with **12 killing tests**, each verified to fail on the bug before passing on the fix.

### Jitter formula (guards the shipped `retry_jitter`)
Tests pin the `_random` seam and capture the actual `asyncio.sleep` argument, asserting `retry_interval * (1 + jitter * (2*r - 1))`. Kills `*`→`/`, `1.0+`→`2.0+`, `2*r-1`→`2*r+1`, and the no-jitter branch, on both `Lock.acquire` (sync thread variant too) and the `LeaderElection` renew loop.

### Graceful-stop release (leadership-leak guard)
`test_graceful_stop_awaits_release_before_returning` gates `backend.release` on an event and proves the renew loop waits for `_release()` before returning. The pre-existing graceful-stop test did not catch the `while not done()`→`while done()` mutation; this one does.

### Error name (cheap, broad)
Each coordination error (`LockReentrantError`, `LockAcquireError`, `LockReleaseError`, `LockNotOwnedError`, `LockLockedCheckError`, `LockOwnedCheckError`, and the LeaderElection `WouldBlockError` guard) is triggered via the real API and asserts the name reaches the message, killing the `name=None` call-site mutations.

Tests only, no production changes. 516 coordination unit tests pass, ty/ruff clean, targeted files stay at 100% coverage.

### Honestly not killed
The `break`→`return` mutation after `sleep_or_stop` is an **equivalent mutation** in Python: a `return` in the `try` still runs the `finally`, so both paths release leadership identically. No test was added that would falsely claim to kill it.

### DIFF
--- a/tests/coordination/test_leaderelection.py
+++ b/tests/coordination/test_leaderelection.py
@@ -718,12 +718,26 @@ async def test_leader_election_stops_gracefully_on_stop_event(
 
 
 async def test_graceful_stop_awaits_release_before_returning(
-    leader_election: LeaderElection,
     backend: LeaderElectionBackend,
     mocker: MockerFixture,
 ) -> None:
     """Graceful stop blocks on `_release()` until the backend release returns."""
-    # Arrange: gate the backend release so it cannot complete until allowed.
+    # Arrange: a generous backend_timeout so the gated release, not the
+    # timeout, decides when `_release` completes.
+    leader_election = LeaderElection.from_config(
+        LEADER_NAME,
+        LeaderElectionConfig(
+            worker="worker_graceful",
+            lease_duration=0.7,
+            renew_deadline=0.6,
+            retry_interval=0.01,
+            error_interval=0.01,
+            backend_timeout=0.5,
+        ),
+        backend=backend,
+    )
+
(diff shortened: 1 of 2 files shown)


grelinfo:
Copilot review triage, the one comment was valid and is fixed in 91e4975: the graceful-stop test built on the shared fixture's `backend_timeout=0.005`, so `_release()` could time out before the gate opened and pass for the wrong reason on slow CI. The test now builds its own LeaderElection with `backend_timeout=0.5` (well above any scheduling delay), so the gated release, not the timeout, drives the assertion. Re-verified it still kills the `while not done()`→`while done()` mutation (fails on the mutation, passes on the fix) and ran it 6x clean.

PR #366, opened 2026-06-15
✨ Add a built-in readiness check per provider with auto_health

grelinfo:
Closes #363.

Every grelmicro app that exposes `/readyz` wants to answer "can I reach my backends?" Today the user hand-rolls one check per provider. The demo proved the friction by probing Redis with a borrowed `Lock("healthz").locked()` round trip, which is neither the cheapest nor the most honest readiness signal for a connection pool.

## What this adds

* `Provider.check()`: an async readiness probe on every connection provider. Redis and Valkey run `PING`, Postgres and SQLite run `SELECT 1`. The base method raises `NotImplementedError`, so a provider with no backend opts out.
* `HealthChecks.add_provider(provider, *, name=None, critical=True, timeout=None)`: registers the probe as a `provider:{short_name}` check. Critical by default, so an unreachable backend fails `/readyz`. Pass `critical=False` for a degradable dependency such as a cache, or `name=` to disambiguate two providers of the same vendor.
* `HealthChecks(auto_health=True)`: off by default. On startup it discovers every active provider (including ones a Component borrows without listing) and registers a critical `provider:{short_name}` check for each. An explicit `add_provider` wins, and re-entry is idempotent. A second provider of the same vendor is skipped with a warning.
* `Grelmicro.providers`: lists active providers, deduped by identity.

## Settled decisions

* `auto_health` defaults to **off** (explicit opt-in, no-magic stance).
* Provider checks default to **critical** (matches the demo they replace and the `/readyz` framing).

## Demo

The FastAPI demo drops its hand-rolled `check_redis` for `HealthChecks(auto_health=True)`, which now probes both Redis and Postgres with no boilerplate.

## Tests and docs

Per-provider `check()` tests, full `add_provider` and `auto_health` coverage, a `Grelmicro.providers` test, a new "Provider Readiness Checks" docs section with runnable snippets, a `providers.md` note, and a changelog entry. The public-API snapshot is updated for the new `auto_health` parameter.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-* ✨ Add a built-in readiness check per provider. Every Provider ships a cheap `check()` probe (Redis and Valkey `PING`, Postgres and SQLite `SELECT 1`). Register it with `health.add_provider(redis)` as a critical `provider:redis` check, or register one for every active provider at once with `HealthChecks(auto_health=True)`. `Grelmicro.providers` lists the active providers.
+* ✨ Add a built-in readiness check per provider. Every connection provider ships a cheap `check()` probe (Redis and Valkey `PING`, Postgres and SQLite `SELECT 1`). Register it with `health.add_provider(redis)` as a critical `provider:redis` check, or register one for every active provider at once with `HealthChecks(auto_health=True)`. `Grelmicro.providers` lists the active providers.
 
 ## 1.0.0a1 - 2026-06-12
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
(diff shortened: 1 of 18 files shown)


grelinfo:
Thanks Copilot. All five comments were valid and are addressed in db23555:

1. **auto_health duplicate check** — `_register_active_providers` now skips a provider whose `check` is already registered under *any* name, so an explicit `add_provider(provider, name='cache', critical=False)` is no longer shadowed by a second critical `provider:{short_name}` check.
2. **from_config** — now accepts and forwards `auto_health`.
3. **providers.md** — softened to "Every connection provider".
4. **changelog** — softened to "Every connection provider".
5. **health.md** — notes that a provider with no built-in probe is skipped.

Added a test for the custom-name case and refreshed the public-API snapshot for the new `from_config` parameter.

PR #370, opened 2026-06-19
📝 Fix verified docs drift across provider matrix, lock cleanup, and stale claims

grelinfo:
## What this fixes

A documentation-accuracy audit cross-checked every doc against the source and found factual drift, all verified against the code. The executed examples (137 snippet/README cases) were already clean; the drift lived in reference tables and prose that no test covers.

### Provider factory matrix (providers.md)
- SQLite `.cache()` and `.circuitbreaker()` cells showed N/A but both are implemented (`sqlite.py:235,243`). Corrected to shipped.
- Prose claiming Postgres "ships `.lock()` and `.leaderelection()`" and SQLite "ships `.lock()` and `.ratelimiter()`" was wrong (both ship more). Corrected to the real factory sets.

### Lock cleanup is vacate-in-place, not DELETE (coordination.md, kubernetes.md)
The docs said expired locks are bulk-DELETEd; the code does `UPDATE ... SET token = NULL, expire_at = NULL` (Postgres/SQLite) and a REPLACE that clears the holder fields (Kubernetes), keeping the row/Lease so the fence counter survives. The old text contradicted the same file's fencing section. Rewritten to match.

### SQLite coverage gaps
- capabilities.md: `LeaderElection × SQLite` was marked shipped, but there is no SQLite election adapter. Now N/A with a legend note.
- comparison.md: SQLite was missing from the Cache and CircuitBreaker backend lists. Added.

### Stale claims
- cache.md: the `@cached` table listed `cache` as required (now optional, mutually exclusive with `ttl=`) and omitted `ttl`/`maxsize` rows. Fixed.
- multiple-apps.md: claimed a second app with `Metrics` raises `MultipleActiveAppsError`; only `Log`/`Trace` are cross-app guarded. Corrected.
- health.md (+ its snippet): non-`HealthError` exceptions render as `"{ExceptionType}: {message}"`, not a generic string. Fixed.
- roadmap.md: removed the "per-provider health check (#363)" bullet, which shipped in 1.0.0a1.
- Smaller fixes: plugins.md (added `coordination.schedule`), metrics.md (up_down_counter not gauge), coordination.md (env prefix `{NAME_UPPER}`, point to `extend()` not `do_acquire`), task.md (`started_at`), and two more `uses=[provider, X(provider)]` redundancies the earlier sweep missed.

mkdocs builds clean, 137 snippet/README tests pass, full gate green.

### DIFF
--- a/docs/architecture/coordination.md
+++ b/docs/architecture/coordination.md
@@ -52,14 +52,14 @@ This prevents accidental collisions between different primitive types sharing th
 Expired locks are never actively removed during normal operation. Instead, all backends use a **lazy filtering** strategy combined with **cleanup on exit**:
 
 1. **Lazy filtering**: Every `locked()`, `owned()`, and `acquire()` call includes an expiry check (`expire_at >= now`), so expired locks are simply ignored without requiring deletion.
-2. **Cleanup on exit**: When the backend context manager exits (`__aexit__`), all expired locks are deleted in bulk. This keeps storage clean across graceful restarts.
+2. **Cleanup on exit**: When the backend context manager exits (`__aexit__`), expired locks are vacated in place. This keeps storage clean across graceful restarts.
 
 If the process crashes without exiting the context manager, expired locks remain in storage but are harmless: they will be filtered out by all subsequent operations and cleaned up on the next graceful shutdown.
 
 ### Backend-specific cleanup
 
-- **SQLite / PostgreSQL**: A single bulk `DELETE ... WHERE expire_at < now` removes all stale rows before closing the connection.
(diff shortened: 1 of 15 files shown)


grelinfo:
Copilot review triage, all valid, fixed:
1. **providers.md factory table missing `.schedule()`** (×3 comments): added the `.schedule()` row (implemented on all four providers) and added `.schedule()` to both the Postgres and SQLite prose lists.
2. **comparison.md LeaderElection backends "same as Lock"**: corrected to `Redis, PostgreSQL, Kubernetes Lease, Memory (no SQLite)` since there is no SQLite leader-election adapter.
3. **cache.md `@cached` maxsize default `128`**: corrected to `0` (unlimited), matching the code.
4. **todo.md**: that was a local scratch file swept in by `git add -A`, now removed from the PR.

PR #372, opened 2026-06-19
♻️ Dedup jitter formula, worker guard, and thread token in coordination internals

grelinfo:
## What this does

Behavior-preserving internal refactors found by a code-quality review. No public API change, coverage stays 100%, the snapshot is unchanged.

- **Jitter formula consolidated.** The 7-line `retry_interval * (1 + jitter*(2*random()-1))` block was copied at three sites (`Lock.acquire`, `Lock.do_thread_acquire`, the `LeaderElection` renew loop), and the `_random` seam was declared in two modules. Extracted to one `jittered_interval(base, jitter)` helper plus one seam in `coordination/_base.py`. The jitter killing-tests were repointed to the new seam and re-verified to still kill a formula mutation (mutate the helper, tests fail; revert, tests pass).
- **`Lock._thread_token` deleted.** It re-implemented the public `generate_thread_token` from `_tokens.py`. Added an explicit `thread_id=` path to that helper and routed the four call sites through it, dropping one `# noqa: SLF001`.
- **Worker-immutability guard extracted.** The identical guard in three `_apply_reconfigure` methods is now one `assert_worker_unchanged(current, new)` call, preserving the exact ValueError message.
- **Removed a spurious `# pragma: no cover`** on `CircuitBreaker.backend` (the branch is exercised by tests, and the five sibling backend properties have no pragma). Coverage stays 100%.

Net effect: ~30 lines of duplication removed, one source of truth for the jitter math (which the mutation review flagged), and the coordination internals read consistently.

Full gate green: 2879 unit + 211 integration, 100% coverage, ruff and ty clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ tmp/
 
 # AI assistant files
 .claude
+todo.md
--- a/grelmicro/coordination/_base.py
+++ b/grelmicro/coordination/_base.py
@@ -1,5 +1,6 @@
 """Lock Base Classes."""
 
+import random
 from types import TracebackType
 from typing import Annotated, Protocol
 from uuid import UUID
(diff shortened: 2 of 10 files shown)


grelinfo:
Copilot triage: the only comment was `todo.md` being swept into the PR by `git add -A` again. Removed it from the PR and added it to `.gitignore` so the local scratch file cannot recur. (The codecov/project red was the usual transient pre-processing state, it now reads 100% +0.00%.)

PR #384, opened 2026-06-21
✅ Fix from_thread test flakiness and the latent CI ty errors

grelinfo:
Fixes the recurring flake in `tests/coordination/test_lock.py::test_lock_from_thread_owned` (and the other from_thread tests) that has been forcing `--no-verify` on commits.

## Root cause
The module sets a blanket `pytest.mark.timeout(1)`. Ten of the from_thread tests do 3-5 sequential thread-to-loop round-trips via `asyncio.run_coroutine_threadsafe(...).result()`. Under heavy parallel CI load the loop thread is descheduled often enough that the accumulated round-trip latency exceeds 1s, and pytest-timeout fails the test. The bridge itself is correct; the timeout was simply too tight for cross-thread work under contention.

## Fix
Raise the module timeout `1s -> 5s`. The timeout guards against genuine hangs (which fail regardless of the value), so the extra headroom removes the flake without weakening that guarantee. `test_lock_from_thread_owned` (the reported offender, 5 round-trips) now passes even under load.

## Note
A separate, pre-existing flake in `tests/providers/test_valkey_provider.py::test_circuitbreaker_factory` surfaced under the same extreme local load (passes in isolation). It is unrelated to this change and worth its own look if CI shows it.

## Test plan
- `test_lock.py` passes in full under `-n auto` locally; the from_thread subset is green.
- Strictly loosens a timeout, so it cannot regress behavior.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -306,9 +306,12 @@ max-args = 10
 [tool.ty.src]
 # The logo build script is an adhoc PEP 723 utility (``uv run`` pulls its
 # deps into an ephemeral env). Project-level type checking can't see
-# ``resvg_py`` or ``fontTools`` and has no reason to.
+# ``resvg_py`` or ``fontTools`` and has no reason to. The mutmut runner
+# imports ``mutmut``, which lives in the optional ``mutation`` dependency
+# group that CI does not install.
 exclude = [
     "docs/img/logo/build_logo.py",
+    "tools/run_mutmut.py",
 ]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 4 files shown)


grelinfo:
Also fixes 4 latent `ty` errors on main (in test_match_validation.py and test_reconfigure_tracking.py) that the full CI Lint job surfaced here. They reached main because earlier stacked admin-merges + `--no-verify` commits bypassed the ty-check. Added the project-standard `# ty: ignore` for the intentionally-invalid / runtime-coerced test arguments.

PR #733, opened 2026-08-15
✨ Report what the app is wired with and default config per kind

grelinfo:
A review of the grelmicro API against Laravel, Symfony, Django, FastAPI, and Spring Boot, then the changes it justified. Every proposal was put to an adversarial pass first, and most of the removals it produced did not survive that pass. What ships here is what did.

## Breaking

The bare `GREL_{PATTERN}_` prefix becomes the default for **every** instance of a pattern, not just the default instance. `GREL_LOCK_LEASE_DURATION=60` used to reach `Lock("default")` alone. It now sets the lease for every `Lock` without its own `GREL_LOCK_{NAME}_LEASE_DURATION`, so a twelve-lock service tunes one variable instead of twelve.

This is the shape Resilience4j uses (`configs.default` + `instances.<name>`), and it is the largest config ergonomics gap the comparison found. It is breaking, so it lands before 1.0.

## Added

- **`micro.describe()` and `python -m grelmicro check`.** One report of what got wired, from what, reachable how far, and configured with what. Credentials masked. The CLI exits non-zero on a failing check so CI can gate on wiring. This closes the gap against `/actuator/beans`, `artisan about`, `debug:container`, and `manage.py check`, all of which the five frameworks ship and grelmicro did not.
- **Providers report the kinds they decline.** `uses=[redis]` leaving the outbox unwired was swallowed as a `NotImplementedError`. It is now the first thing the report shows, which is grelmicro's `/actuator/conditions`.
- **`grelmicro.integrations` entry-point group** replaces `hasattr` sniffing in `install()`. Third parties can ship an integration, and a `FastAPI` subclass in your own package now resolves because the lookup walks the MRO.
- **`micro.fake()`** swaps every backed component onto an in-process store for a block, so a fixture can open the app you actually ship and fake only its backends.
- **Typed `micro.get(Cache)`.** Additive: `__getattr__` and the string form are untouched.
- **Diagnostic codes.** Every startup report carries a stable slug (`backend-scope`), trailing the message, travelling as a `diagnostic` log field, linked to a new reference page. Each warning gets its own category under `GrelmicroConfigWarning`, so one diagnostic is silenced by category rather than by matching message text, which is what that class already promised.

## Fixed

- **A bare backend matching two protocols resolved by `if` statement order.** `runtime_checkable` compares member names only, so every `CircuitBreakerBackend` also matched `RateLimiterBackend`. Swapping two lines would have silently registered every breaker as a rate limiter. Most-specific-wins now decides it, and a genuine tie raises `AmbiguousBackendError`. A test pins the resolved kind for every first-party adapter.
- **`instantiate_if_class` misreported a `TypeError` from inside `__init__`** as "needs constructor arguments", suggesting a fix that could not work. The arity check now reads the signature before calling.

## Not done, deliberately

Seven proposed removals were refuted on evidence and are recorded as such rather than attempted: the three ordering passes, the bare-backend auto-wrap, `instantiate_if_class` itself, the ambient bundle, `MultipleActiveAppsError`, `strict=`, the shield config exports, and `__getattr__`. Pruning the error surface was also dropped: all 63 exported errors turn out to be documented and tested, so the cross-framework argument for it did not hold.

## Verification

Full tier green: unit + integration + slow, 100% coverage, ruff, ty, mypy, and `mkdocs build --strict`.


grelinfo:
Thanks, all three were valid. Each verified against the code and fixed in 8d0220d:

1. **Duplicated constants in `tests/test_config.py`** — real. A scripted edit inserted the block twice. Ruff did not catch it because reassigning a module-level constant is not `F811` (that rule covers imports and functions), so only a reader would. Second block removed.
2. **`is_bound(app, micro)` in the `INTEGRATION_GROUP` docstring** — real mismatch. The protocol, both first-party integrations, and `docs/architecture/plugins.md` all take `is_bound(app)`. The docstring was the only wrong one, and it is the one a third-party author would read first. Corrected.
3. **Stale `/diag/<code>` in the sample report** — real. The URL shape changed to `/diagnostics/#<code>` late (an anchor is safe here because the heading *is* the code), and this sample was not updated with it. Fixed, and I checked the sample now matches `diagnostic()` output exactly rather than trusting the edit.

PR #734, opened 2026-08-15
🐛 Reach every reconfigurable component and keep first-party types

grelinfo:
Follow-up to the 0.38.0 API review. An independent review of the public API (GitHub Copilot CLI, `gpt-5.6-sol`, high effort) produced 50 findings. Every one was put through an adversarial verification pass before anything was changed, and a large share did not survive it.

## What the verification rejected

Of 9 CRITICAL findings, **4 were wrong** and would each have broken documented, tested behaviour:

- **Scoping `_reconfigurables` per app** is not implementable as proposed. The set tracks *patterns* (`Lock`, `RateLimiter`), which are constructed at module load before any app exists. That is the documented primary use case in `ExternalConfig`'s own docstring and in the ConfigMap guide.
- **"Provider defaults behave inconsistently"** is deliberate and documented, and there is a test whose comment refutes the finding verbatim. The behaviour is also load-bearing for `HealthChecks(auto_health=True)`.
- **Renaming `ratelimiter` to `rate_limiter`** confuses two deliberate naming domains: concatenated identifiers for Python surfaces (kinds, env prefixes, module paths), snake_case for OpenTelemetry metric names. They never meet in one lookup. The rename would churn 100+ sites including env namespaces already held in user ConfigMaps.
- **"No stable construction grammar"** — the grammar is documented as a three-category contract, and the proposed single form cannot express variant-driven patterns like `RateLimiter`.

Of 33 MAJOR findings, 11 were rejected. Notably **all seven removal or demotion proposals turned out to be load-bearing** or documented deliberate design.

## Fixed

**Real bugs**

- `HealthChecks` inherited `Reconfigurable` but never registered itself, so `ExternalConfig` could never retune it from a mounted ConfigMap. It was the only reconfigurable component that did not track itself. Verified safe: the config is frozen with validated fields, and a running probe snapshots its config, so a mid-round patch cannot break it.
- `TTLCache.get_or_set` and `Idempotency.run` awaited only coroutines while their annotations promise `Awaitable`, so a factory returning a `Future` had the **awaitable itself stored** as the cached value or the idempotent response.
- `TrustedProxies` accepted `max_entries=0`, which slices as `entries[-0:]` and returns the whole list, so a cap meaning "allow nothing" silently allowed everything.
- The singleton guard inspected only the incoming component, so a plain component of a singleton kind could register after the singleton.
- `CallLog.count(name=None)` matched calls that never passed `name`.
- `ExternalConfig(reload_interval=0)` polled without pause.

**Typing**

- `@idempotent` accepted a synchronous function that then failed at runtime, and erased the decorated return type. It now requires a coroutine function and preserves the type.
- Typed properties for `health`, `metrics`, `outbox`, `ratelimiter`, `circuitbreaker`. Every first-party kind keeps its type through lookup.
- `Metrics.provider`, `Metrics.prometheus_registry`, `Trace.provider` return their real OTel and Prometheus types.

**Exports**

- `grelmicro.describe` gives the `describe()` report models a public home. `import grelmicro` still does not pay for them, pinned by a test.
- `LockConfig`, `TaskLockConfig`, `Task`, and the four `Log*Type` enums were all required by public signatures but unexported.

## Two deliberate deviations from the review's advice

- The review wanted `key: Callable[P, str]` on `@idempotent`. That breaks the documented `lambda **kw: kw["idempotency_key"]` form in `docs/idempotency/index.md`, because the lambda binds the signature the function must then match. Kept `Callable[..., str]`, which still fixes both real defects.
- Typing `prometheus_registry` surfaced a narrowing gap in `metrics/fastapi.py`. Not reachable today, but the invariant is now stated rather than silenced.

## Not done

Seven findings are real but need a design decision rather than a patch, and are deliberately left out: `Outbox.shutdown_timeout` placement, Fallback/Shield generics, `Outcome` as a discriminated union, `TaskRouter` method naming, `Coordination` factory signatures, provider config strictness, and the validation-error contract (where both proposed directions break something).

## Verification

4487 unit plus 296 integration and slow tests pass, 100% coverage, ruff, ty, mypy, and `mkdocs build --strict` all clean.


grelinfo:
Thanks. Two of the four are one finding, and it was a good catch. Verdicts:

**`Idempotency[R]` erases the return type (comments 3 and 4): CORRECT, fixed in 1379aa4.**

Proven rather than assumed. With the documented `idem = Idempotency("charge")` form, mypy reports:

```
error: Expression is of type "Any", not "int"  [assert-type]
```

So binding the parameter to `R` made the common case solve `R` as `Any`, which is the opposite of what the change was for. Applied your fix exactly: `Idempotency[Any]` on the parameter, `cast("R", ...)` on the awaited result, so `R` is bound solely by the decorated function's own annotation. All three properties now hold, each checked:

- unparameterized `Idempotency("charge")` -> returns `int`
- parameterized `Idempotency[int]` -> returns `int`
- a sync function is still a type error rather than a runtime failure

Added a type assertion in `tests/typechecking/app.py` pinning the unparameterized form, since that is the one that silently regressed.

**Optional-dependency TYPE_CHECKING imports (comments 1 and 2): valid concern, but not introduced here, so out of scope for this PR.**

The pattern predates the change. `grelmicro/metrics/_component.py` already imported `Counter`, `Histogram`, `Meter`, and `UpDownCounter` from `opentelemetry.metrics` under `TYPE_CHECKING`, and `opentelemetry-api` is equally an extra rather than a base dependency (`pyproject.toml:157`). My additions of `opentelemetry.sdk` and `prometheus_client` are the same class of import, so they add no failure mode that was not already there.

The project also already configures for it: `ignore_missing_imports = true` (`pyproject.toml:408`).

If the base install should be type-checkable without extras, that is a real and separate question covering every optional import in the package, not just these three lines. Worth an issue rather than a change smuggled into this PR.

Issue #758, opened 2026-08-16
✅ Replace the hand-rolled public API snapshot with a tool that derives the surface


grelinfo:
`tests/test_public_api.py` is a hand-rolled API freeze guard: a `PUBLIC_MODULES`
list, a signature normalizer, and a syrupy JSON snapshot. It works, but the list
is maintained by hand, so it only guards what someone remembered to add.

It did drift. `grelmicro.outbox` and `grelmicro.types` were rendered in
`docs/reference/` for releases while their `__all__` sat outside the snapshot, so
a rename in either could have shipped unnoticed. [#750](https://github.com/grelinfo/grelmicro/issues/750)
added them plus a test that reads the reference pages, which closes that one hole
by deriving the list from the docs instead of from memory.

That fix is still the same shape as the bug: our own code, guarding our own list.
The surface should come from the package.

## What to find out

Whether a maintained tool already does this better than we do, before 1.0 freezes
the API and the guard starts mattering.

- **[griffe](https://mkdocstrings.github.io/griffe/)** is the strongest lead. It
  is the library mkdocstrings already uses to build our reference pages, so it is
  effectively a dependency we ship docs with. `griffe check` diffs a package
  against a previous version or a git ref and reports breaking changes
  (removed/renamed parameters, changed kinds, dropped exports) with a
  `--verbose` explanation per finding.
- **[mypy stubtest](https://mypy.readthedocs.io/en/stable/stubtest.html)** solves
  a neighbouring problem (stubs versus runtime), so check what it catches here.
- Anything else current. `cargo-semver-checks` and `api-extractor` are the bar in
  other ecosystems, so name the Python equivalent if one has taken hold.

## What the answer has to cover

- Does it discover the surface itself, or does it need a list too? A tool with
  its own list is not an improvement.
- Does it catch what ours catches: parameter rename, reorder, removal, and a
  required-to-optional flip? Our snapshot deliberately drops annotation reprs and
  default values because they vary across the supported Python and Pydantic
  matrix. Does the tool stay stable across that matrix, or does it go flaky where
  we went conservative?
- Does it read `__all__`, or does it treat every non-underscore name as public?
  Those differ a lot here.
- Does it run offline in pre-commit and CI, against a git ref rather than a
  published release? Release-to-release only is not enough, the guard has to fire
  on the PR.
- Is the output reviewable in a diff, the way the committed JSON snapshot is?

## Done means

Either the snapshot is replaced by the tool, or the issue is closed with the
[...]

PR #759, opened 2026-08-16
💥 Raise one SettingsValidationError for every bad configuration value

### LINKED ISSUE #750: 📝 Decide whether a config validation error is wrapped or raw
A bad configuration value raises two different exception types depending on which class you built.

Components wrap. `Outbox`, `Metrics`, `Trace`, `HealthChecks`, `Idempotency`, `Log`, and `Tasks` all pass `error_type=` to `resolve_config`, so a bad value surfaces as their own `*SettingsValidationError`.

Patterns do not. `Lock`, `Timeout`, `Retry`, `Bulkhead`, `Fallback`, `Shield`, `RateLimiter`, and `CircuitBreaker` let the raw `pydantic.ValidationError` propagate.

Nothing in the docs says which to expect, so a caller writing `except` has to know which half of the library they are holding.

**The argument for wrapping everything.** `SettingsValidationError` deliberately omits the offending value: settings come from environment variables that may carry credentials, and echoing the value into a log line leaks them. That reason applies to patterns exactly as much as to components, and it got stronger now that `RateLimiter` and `CircuitBreaker` read the environment too.

**The argument for raw everywhere.** Pydantic's message already describes the failure precisely, and one fewer wrapper is one fewer thing to learn.

Either answer is defensible. Two answers is not.

## Architecture constraints

**Whichever way this goes, it goes the same way everywhere.** The point of the change is removing the split, so a partial fix that converts two classes and leaves six is worse than leaving it alone.

**Do not leak the offending value.** If the outcome is raw `ValidationError`, confirm that pydantic's rendering cannot echo a credential read from the environment, and say so in the docs. If it cannot be confirmed, wrap.

**This is a clean cut.** Pre-1.0, so change the type and put the entry under `### Breaking`. No alias, no shim.

**State the answer in the docs.** [Configuration internals](https://grelmicro.grel.info/architecture/config/) carries the environment contract. The exception a bad value raises belongs there, next to R7, which already says a bad value is reported with a validation error naming the field.

grelinfo:
Closes #750.

A bad configuration value raised three different exception types depending on
which class you built. Now it raises one, everywhere.

## The decision

`SettingsValidationError`, for every pattern and every component. The ten
per-module subclasses are deleted, not deprecated, which is the pre-1.0 rule.

Two things settled it. No caller ever reacted differently by module: there is
not one `except *SettingsValidationError` in `docs/`, `examples/`, or the
README, and none of `except CoordinationError` / `except ResilienceError`
either. And the message already names the exact variable, which locates the
problem better than a class name can. R7 in
[Configuration internals](https://grelmicro.grel.info/architecture/config/)
and `docs/migration.md` were already written in terms of the base error, so the
code was contradicting the published contract rather than extending it.

Removed: `CacheSettingsValidationError`, `CoordinationSettingsValidationError`,
`HealthSettingsValidationError`, `IdempotencySettingsValidationError`,
`LogSettingsValidationError`, `MetricsSettingsValidationError`,
`OutboxSettingsValidationError`, `ResilienceSettingsValidationError`,
`TaskSettingsValidationError`, `TraceSettingsValidationError`.

`except SettingsValidationError`, `except ValueError`, and
`except GrelmicroError` all keep working.

## Two defects found on the way, neither in the issue

**0.39.0 shipped a false claim.** Its release notes say "Every configuration
path raises `SettingsValidationError` for a bad value". `Fallback`, `Shield`,
and `TTLCache` resolve their own configuration instead of going through
`resolve_config`, so all three still let pydantic's error through, carrying the
rejected value:

```
GREL_FALLBACK_F_WHEN=secret  ->  ValidationError: ... input_value='secret'
GREL_SHIELD_S_MAX_RATE=token ->  ValueError: could not convert string to float: 'token'
```

`Shield` was the worst of the three: a bare `ValueError` out of an unguarded
`float()`, which pydantic never saw, so no pydantic setting could have fixed it.
That changelog entry is corrected in place rather than left standing.

**Wrapping alone does not satisfy R7.** This is the one worth reading. The
wrapper strips pydantic's `input_value`, but it renders `msg` verbatim, so a
validator that interpolates the rejected value into its own message leaks
straight through the wrapper:

```
- GREL_FALLBACK_F_WHEN: Value error, when= env entry must be a fully-qualified name, got 'secret-value-here'
```

Shipping only the type change would have repeated 0.39.0's mistake: claiming a
leak was closed while it was open. The `when=` and `timeout_errors` messages on
`Retry`, `Fallback`, and `Shield` now name the module path and the attribute
instead of the raw entry.

R7 is amended to state the rule the code can actually keep: pydantic's automatic
echo is always stripped, and a validator may name a value only where the field's
domain is a public closed set and the value is the whole diagnostic, as an
unknown timezone name is. That is a per-field decision, never the default.

## Also in here

- `error_type` is removed from `resolve_config` and `resolve_config_from_mapping`.
  On the mapping function it was dead: the only caller, `reconfigure_all`, never
  passed it and deliberately catches the raw error to log a redacted summary.
- New `docs/reference/errors.md`. `SettingsValidationError` is now the single
  configuration error and had no documented home.
- The public API snapshot covers `grelmicro.outbox` and `grelmicro.types`. Both
  were rendered in the API reference while their `__all__` went unguarded. A new
  test reads the reference pages, so documenting a module now guards it.
  Replacing the whole hand-rolled snapshot with a real tool is #758.

## Scope I did not take

`RedisProviderConfigError` and `PostgresProviderConfigError` stay. Providers read
the vendor namespace under R1 (`REDIS_*`, `POSTGRES_*`), not `GREL_*`, so those
are per-problem rather than per-module, and `docs/providers/index.md` documents
them as a pair.

## Verification

`tests/test_settings_error_contract.py` sweeps every pattern on both the env and
the kwargs path, asserting the type and asserting the rejected value is absent
from the message. It has a floor so a trimmed list cannot pass silently.

I checked it fails when the bug returns, rather than trusting it because it is
green: reintroducing the 0.39.0 `Fallback` behaviour fails both the type and the
leak assertion.

Full `pre-commit run --all-files` passes, including the integration tier and the
100% coverage gate. Every file touched here is at 100%.

### DIFF
--- a/docs/architecture/config.md
+++ b/docs/architecture/config.md
@@ -86,6 +86,13 @@ in any of its algorithms, is always accounted for:
 | Live reload, field of another algorithm | Skipped for that instance and counted at debug, never a crash. Key names stay out of the logs because a mounted Secret's key can itself be sensitive. A warning naming sibling fields follows with config provenance ([#664](https://github.com/grelinfo/grelmicro/issues/664)). |
 | Value invalid | `SettingsValidationError` naming the variable and the reason, never the value |
 
+Every class raises that one error, whichever pattern or component it is. There
+is no per-module subclass to remember: a bad value is a startup crash, and the
+message already names the exact variable, which locates the problem better than
+a class name can. A config class you build yourself, such as `RetryConfig(...)`
+or `ExponentialBackoff(...)`, raises pydantic's `ValidationError` like any
+pydantic model.
+
 For the default instance the instance address is the kind address, so it
 follows the broadcast row: the bare prefix cannot tell "this instance's field"
 from "every instance's field", and an ambiguous address is not an unambiguous
@@ -94,6 +101,13 @@ mistake.
 The rejected value is never echoed. A variable name is the operator's own,
 but its value can be a credential, so an error carries the name and the reason
 and stops there.
+
+Pydantic attaches the rejected input to every error it raises, for every field,
+whether or not anyone reviewed that field for secrets. Wrapping into
+`SettingsValidationError` is what removes it. A validator may still name a
+value on purpose where the field's domain is a public closed set and the value
+is the whole diagnostic, as an unknown timezone name does. That is a per-field
+decision, never the default.
 
 A name matching no declared field of any algorithm is ignored without report.
 Kubernetes injects `{SVCNAME}_SERVICE_HOST` into every pod, so grelmicro must
@@ -121,6 +135,7 @@ not.
 | `from_config` is the one door for a pre-built config | The environment-merging lane and the config-is-truth lane must be distinguishable at the call site | The environment lane is removed |
 | `GREL_SHIELD_PROFILE` selects a preset, not an algorithm | Every profile declares the identical field names, so no variable gains or loses meaning from the choice | A profile adds or removes a field |
 | A Provider reads its vendor namespace, not `GREL_*` | Connection settings belong to the deployment, and every vendor already defines those names | grelmicro starts owning connection settings |
+| One `SettingsValidationError` for every class, no per-module subclass | No caller reacts differently to a bad value by module, and the message names the variable | A caller needs to branch on the module a config error came from |
 
 ## `resolve_config()`
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,10 +1,27 @@
 # Changelog
 
+## Unreleased
+
+### Breaking
+
+* 💥 A bad configuration value raises `SettingsValidationError`, whichever pattern or component you built. The ten per-module subclasses are removed: `CacheSettingsValidationError`, `CoordinationSettingsValidationError`, `HealthSettingsValidationError`, `IdempotencySettingsValidationError`, `LogSettingsValidationError`, `MetricsSettingsValidationError`, `OutboxSettingsValidationError`, `ResilienceSettingsValidationError`, `TaskSettingsValidationError`, and `TraceSettingsValidationError`. Catch the base error, which every one of them already subclassed. No caller ever reacted differently by module, and the message names the exact variable. ([#750](https://github.com/grelinfo/grelmicro/issues/750))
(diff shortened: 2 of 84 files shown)


grelinfo:
Review round done. Everything Copilot raised and everything my own review found is fixed in 14cc3e7a.

**Copilot, 8 comments, all valid and all applied.** Five tautological
`issubclass(SettingsValidationError, SettingsValidationError)` tests deleted (the
test above each already asserts the real contract). `tests/cache/test_ttl.py`
docstrings corrected. `_fill_from_env`'s message now names the variable rather
than the field, matching its docstring and the shape `resolve_config` renders.
`read_text(encoding="utf-8")` pinned.

**Two substantive findings from my own pass**, both verified before and after:

*A bare `TypeError` escaped the contract.* Pydantic converts only `ValueError`
and `AssertionError` into a `ValidationError`, so a `when=` entry resolving to a
non-exception escaped `except SettingsValidationError` and `except ValueError`
alike:

```
GREL_ENV_LOAD=1 GREL_FALLBACK_F_WHEN=os.getcwd
  -> TypeError: ... which is not an Exception subclass
```

Now `ValueError`, so pydantic wraps it. `ruff`'s `TRY004` wants `TypeError` here,
which is exactly wrong, so those three lines carry a `# noqa: TRY004` with the
reason.

*The message scrub was cosmetic.* It reported `module 'acme.tokens'` plus
`attribute 'Sk_live_abc'`, which together rebuild the rejected value. The
messages no longer name any part of it.

The sweep now covers this: `SECRET` has no dot, so every FQN field rejected it on
the first branch and three later branches went untested. `FQN_CASES` adds a
dotted value and an importable-but-not-an-exception value across all three
fields.

**Not in this PR:** an audit turned up ten more places where the contract is
still broken outside the config-resolution path, filed as #760. The worst is not
a taxonomy problem at all: `Match.exception` raises bare `TypeError` and
`reconfigure_all` guards only `except ValidationError`, so one bad key in a
mounted ConfigMap takes down the entire resync cycle. That is pre-existing and
independent of this change, which is why it is a separate issue rather than
scope creep here.

#760 also argues the R7 carve-out this PR introduced should be removed. The
argument is good and I think it wins, but it changes a rule this PR only just
wrote down, so it belongs in its own change.

PR #761, opened 2026-08-17
🔒 Close every remaining path that leaked a value or escaped the config error

### LINKED ISSUE #760: 🔒 Close the remaining paths that break the one-config-error contract
[#750](https://github.com/grelinfo/grelmicro/issues/750) settled the rule: a bad
configuration value raises `SettingsValidationError` and never carries the value.
[#759](https://github.com/grelinfo/grelmicro/pull/759) landed it for the config
resolution path. An audit afterwards found the rule still broken in these places.

Each one below was reproduced, not read off the source.

## Fix first, this one is a crash not a taxonomy problem

**1. A bad key in a mounted ConfigMap kills the whole resync cycle.**
`Match.exception` raises bare `TypeError` (`grelmicro/resilience/_match.py:143`,
`:272`), and `reconfigure_all` guards only `except ValidationError`
(`grelmicro/_config.py:855`). Pydantic does not convert a validator's
`TypeError`, so it escapes `reconfigure_all` instead of being logged and skipped,
and every other instance in that cycle stops updating.

```bash
GREL_ENV_LOAD=1 GREL_RETRY_RJ_WHEN='[123]' uv run python -c \
  "from grelmicro.resilience import Retry; Retry('rj')"
# TypeError: Match.exception() arguments must all be exception classes, got 123
```

The same shape at `grelmicro/resilience/shield/_profile.py:150-155`:
`Shield("s", timeout_errors=KeyboardInterrupt)` raises bare `TypeError` naming
the class. `#759` fixed exactly this pattern in the three FQN resolvers by
raising `ValueError`, which pydantic does convert. These two were missed.

## Value still reaches the error, from sources grelmicro does not write

`SettingsValidationError` renders pydantic's `msg` verbatim, so any message
pydantic builds with the input in it leaks through the wrapper.

**2. Discriminated union tag.** `GREL_RETRY_{NAME}_BACKOFF='{"kind": "s3cret"}'`
gives `Input tag 's3cret' found using 'kind' does not match any of the expected
tags`. Reachable through `RetryConfig.backoff`.

**3. `ImportString`.** `GREL_CIRCUITBREAKER_{NAME}_IGNORE_EXCEPTIONS='s3cret.Boom'`
gives `Invalid python path: No module named 's3cret'`
(`grelmicro/resilience/circuitbreaker/consecutive_count.py:51`). Same class of
leak the FQN resolvers had, except pydantic supplies the message.

**4. `RandomBackoff._check_range`** (`grelmicro/resilience/backoffs/random.py:56`)
interpolates both delays, and arrives via `GREL_RETRY_{NAME}_BACKOFF`.

## Wrong exception type

**5. Dict-valued component fields raise `pydantic_settings.SettingsError`.**
pydantic-settings fails complex-field JSON decode at the source stage, before
validation, so `resolve_config`'s `except ValidationError` never sees it. Hits
`MetricsConfig.headers` / `.resource_attributes` and the `TraceConfig`
equivalents. It surfaces at app open, so a deployment sees it and the test suite
does not:

```bash
GREL_ENV_LOAD=1 GREL_METRICS_HEADERS=notjson uv run python -c \
  "import asyncio; from grelmicro import Grelmicro; from grelmicro.metrics import Metrics; \
   asyncio.run(Grelmicro(uses=[Metrics()]).__aenter__())"
# pydantic_settings.exceptions.SettingsError: error parsing value for field "headers"
```

**6. `cached()` is split-brained.** `cached(ttl=-1)` raises
`SettingsValidationError`, while `lock=`, `early=`, and `stale_ttl=` raise bare
`ValueError` echoing the value (`grelmicro/cache/cached.py:537,540,543`). One
call, two contracts.

**7. `TrustedProxies`** (`grelmicro/security/clientip.py:260,287,299,306`) raises
bare `ValueError` and `TypeError` echoing the entry.

**8. `ExternalConfig(reload_interval=...)`** (`grelmicro/config/_external.py:167`)
raises bare `ValueError` echoing the value.

## Accepted in silence

**9. `Grelmicro(environment="nonsense")`** is stored verbatim with no error and no
warning (`grelmicro/_environment.py:142`), and the backend-scope check then
treats production as undeclared. The env path warns for the same value, so the
two doors disagree.

**10. `Trace(instrument=...)` is validated only when the exporter is active**
(`grelmicro/_app.py:1086`). A bad directive opens cleanly in development and
raises on the first deploy that sets an endpoint.

## The R7 carve-out should probably go

`#759` added a carve-out to R7: a validator may name a rejected value where the
field's domain is a public closed set, which is what an unknown timezone name
does. The argument against it is strong. The carve-out describes the *accepted*
values, but the echoed string is by definition outside that set, which is exactly
the arbitrary operator input the rule exists to contain. `GREL_TIMEZONE` is an
app-wide alias every `Log` and `Tasks` reads, so a swapped variable echoes
whatever was put there.

The diagnostic survives without the echo: name a *member* of the set instead
("did you mean `Europe/Zurich`"), chosen by closest match. That reads better for
the typo case and a member name can never be a secret.

Related: `UnknownEnvironmentWarning` echoes `GREL_ENVIRONMENT`
(`grelmicro/_environment.py:159`) on both the `warnings` and the logging channel.
The carve-out as written covers validators, not warning channels, so this one is
not covered by anything today.

## Done means

The sweep in `tests/test_settings_error_contract.py` grows to cover these, and
adding a new one fails CI. That test only reaches patterns today, so extend it to
components and to the non-pattern surfaces above.

## Architecture constraints

**Raise `ValueError`, never `TypeError`, from a pydantic validator.** Pydantic
converts only `ValueError` and `AssertionError` into a `ValidationError`.
A `TypeError` escapes every documented `except`, so `ruff`'s `TRY004` is wrong
here and wants a `# noqa: TRY004` with the reason.

**A pydantic model the caller builds directly keeps raising `ValidationError`.**
`RetryConfig(...)` and friends are pydantic models and behave like them. The
contract covers classes that *resolve* configuration.

**Provider errors stay.** `RedisProviderConfigError` and
`PostgresProviderConfigError` cover the vendor namespace under R1, which is a
different contract.

grelinfo:
Closes #760.

#759 made `SettingsValidationError` the one error for a bad configuration value
on the resolution path. An audit afterwards found the rule still broken in
twelve places. This closes all of them.

## The crash, which was not a taxonomy problem

`Match.exception` raised bare `TypeError`, and `reconfigure_all` guarded only
`except ValidationError`. Pydantic never converts a `TypeError`, so it escaped
the reload loop and aborted the whole resync cycle: one bad key in a mounted
ConfigMap stopped every other instance from updating. An empty
`GREL_RETRY_{NAME}_WHEN` was enough to reach it.

The loop now skips any value a validator refuses, whatever it raises, and logs
the class name only, since the message is not ours and may carry the value.

## Three mechanisms, because one is not enough

The audit showed the wrapper alone cannot hold rule R7, so the rule now rests on
three independent things:

1. **Wrapping** drops the input pydantic attaches to every error it raises.
2. **Our validators** are written not to name the value.
3. **A backstop in `SettingsValidationError`** scrubs the input out of the
   message, including the fragments it splits into.

The backstop is what catches messages we do not write. Pydantic puts the input
into some of its own text, and both were reachable from the environment:

```
GREL_RETRY_{NAME}_BACKOFF='{"kind": "s3cret"}'
  -> Input tag '[redacted]' found using 'kind' does not match ...
GREL_CIRCUITBREAKER_{NAME}_IGNORE_EXCEPTIONS='s3cret.Boom'
  -> Invalid python path: No module named '[redacted]'
```

## Wrong type

`GREL_METRICS_HEADERS=notjson` raised `pydantic_settings.SettingsError`, not
`SettingsValidationError`. pydantic-settings JSON-decodes a complex field in the
source stage, before validation, so it never reached `except ValidationError`.
It surfaced at app open, which means a deployment saw it and the test suite did
not. Same for `resource_attributes`, and for both fields on `Trace`.

`cached()` was split-brained: `ttl=-1` raised `SettingsValidationError` while
`lock=`, `early=`, and `stale_ttl=` raised a bare `ValueError` echoing the value.
One call, two contracts. `TrustedProxies` and `ExternalConfig` did the same.

## Accepted in silence

`Grelmicro(environment="prod")` was stored verbatim, and the backend check then
ran as if no tier were declared, which is the check that refuses a memory backend
in production. The environment variable already warned, so the two doors
disagreed. It now raises.

`Trace(instrument=...)` validated the directive only when an exporter was active,
so a bad one opened cleanly in development and first raised on the deploy that
set an endpoint.

## The R7 carve-out from #759 is removed

#759 allowed a validator to name a rejected value where the field's domain is a
public closed set, as an unknown timezone name did. The argument inverts on
contact: the domain describes the values that are *accepted*, while the echoed
string is by definition one that was not. `GREL_TIMEZONE` is an app-wide variable
every `Log` and `Tasks` reads, so a swapped variable puts a credential there.

The message names a *member* of the set instead, which diagnoses the typo better
than repeating it did and can never be a secret:

```
unknown timezone name, did you mean 'Europe/Zurich'
```

`UnknownEnvironmentWarning` echoed `GREL_ENVIRONMENT` on both the warning and the
logging channel, which the carve-out did not even cover. Scrubbed too.

## A rule worth stating

Ruff's `TRY004` asks for `TypeError` on a type check. Inside a validator that is
exactly wrong, because pydantic converts only `ValueError` and `AssertionError`.
Those sites carry a `# noqa: TRY004` with the reason, and
`docs/architecture/config.md` now records it as settled.

## Verification

`tests/test_settings_error_contract.py` grows from 23 to 48 cases: components
driven to real resolution through an open app, the non-pattern surfaces, the FQN
branches, empty values, and the reload loop.

I checked the new tests fail when the bugs return rather than trusting them
because they are green. Reverting the `SettingsError` guard fails three cases.
Reverting `Match` to `TypeError` *and* removing the reload guard fails the resync
test, while reverting `Match` alone does not, which is what proves the guard is
load-bearing rather than decorative.

Full `pre-commit run --all-files` passes: all 17 hooks, including the integration
tier and the 100% coverage gate.


grelinfo:
Both reviews triaged. Copilot raised 3 comments, my own pass raised 9. All valid, all fixed.

**Copilot: three stale `Raises:` sections** on `TrustedProxies`, `ExternalConfig`, and `cached()`, which still documented `ValueError` after the doors moved to `SettingsValidationError`. Fixed, and I swept the other changed files for the same thing rather than only the three flagged. `_timezone.py` keeps its `ValueError` because that one is still accurate: it raises into a pydantic validator, which converts it.

**The `_scrub` backstop was wrong in three ways, all found independently by both the review and my own testing.** It replaced every separator-split fragment of the input, so it destroyed exactly the message an operator needs:

```
GREL_LOG_LEVEL=INF
  -> Input should be 'DEBUG', '[redacted]O', 'WARNING', 'ERROR' or 'CRITICAL'
GREL_TIMEZONE=Europe/Zurichh
  -> unknown timezone name, did you mean '[redacted]/Zurich'
```

The second is worse than cosmetic: the `did you mean` hint is the feature this PR added to replace the echo, and the scrub ate it in the only case it ever fires. A short fragment could also rewrite the placeholder already inserted, giving `[[redacted]acted]`.

Redesigned against what pydantic actually does, checked rather than assumed:

| Error type | Embeds input | Handling |
|---|---|---|
| `import_error` | yes, the module | message replaced outright |
| `union_tag_invalid` | yes, the tag | whole-token replace |
| `value_error`, `assertion_error` | someone else's text | whole-token replace |
| `int_parsing`, `literal_error`, `greater_than`, ... | no | untouched |

Whole-token and single-pass, so a typo that is a prefix of the correct value no longer takes the correct value out of the message offering it.

**A fourth `TypeError` escape.** `_normalize_timeout_errors` still raised `TypeError` with `{value!r}`, so `Shield("x", timeout_errors=int)` escaped both `except SettingsValidationError` and `except ValueError` — the failure mode this PR documents as impossible.

**A regression I introduced.** Validating the `instrument` directive on the inactive path checked names against `installed_instrumentors()`, which reads entry points. An app declaring `Trace(instrument=["sqlalchemy"])` would have started failing in any dev or CI image without the trace extras, while being correct in production, and that branch runs in exactly those images. `validate_directive` now takes `check_names=False`: the map-value check still runs, since it depends on nothing outside the directive, and the names are left to the active path where the set is real.

Also: `Match.exception` reported `got type` for every class, since `type(cls).__name__` is always `"type"`. And the `environment=` docstrings still described the lenient behaviour this PR removes.

Full `pre-commit run --all-files` green, including the integration tier and the 100% coverage gate.

PR #762, opened 2026-08-17
✅ Sweep the backend, configuration, and construction contracts by discovery

### LINKED ISSUE #755: ✅ Sweep every published contract across every implementation
## The problem, stated precisely

The 0.39.0 pre-release review found roughly ten pre-existing defects by reading the code. **The test suite has a 100% coverage gate and caught none of them.** Several affected every pattern in the library:

- `reconfigure` refused the config class its own docs told you to build, for any instance built from the environment.
- A `GREL_{KIND}_{FIELD}` variable set with the gate off was silently dropped on every named instance.
- `SettingsValidationError` raised `IndexError` instead of rendering, for any failure from a model-level validator.
- `MemoryCircuitBreakerAdapter.bind` ran any algorithm as consecutive-count while the other three adapters refused.
- Every rate limiter adapter read the provider client before checking the algorithm, so an unsupported kind needed a live connection to fail.
- `Shield` sat outside the configuration rules the docs published for every pattern.

Coverage measures lines executed, not behaviours verified. Every one of these is a **rule that holds for a whole family of classes**, and every one was tested, if at all, on one member of the family.

## The shape of the fix, which already exists here

`tests/test_adapter_contracts.py` is the right idea. It walks the package with `ast`, finds every class declaring a given protocol, refuses to pass on an empty scan (`_MIN_ADAPTERS`), and parametrizes over what it finds.

It asserts **one** invariant: that adapters capture the running loop.

So the machinery is built and used once. The work is to drive it from the contracts we have now published, so that a rule stated in the docs is a rule enforced against every implementation, including ones added later.

## What to sweep

Each of these is a rule already written in `docs/architecture/`. Each maps to a family the `ast` sweep can enumerate.

**Backend contract** ([Plugins](https://grelmicro.grel.info/architecture/plugins/)), over every backend adapter:

- An unsupported algorithm kind raises `NotImplementedError` naming the kind. Partially done for rate limiters in `test_ratelimiter_unknown_kind.py`, and the circuit breaker gap is exactly what that test's absence hid.
- `bind` dispatches on the kind before touching the provider client, so an unsupported kind fails with no connection. Assert with a provider stub that raises on attribute access.
- `scope` is declared.

**Configuration contract** ([Configuration internals](https://grelmicro.grel.info/architecture/config/)), over every pattern and component:

- R7's table, all five rows, for every class that reads the environment. The kind-address bug was one row failing for every named instance.
- R8: `from_config` reads no variable and does not register for live reload.
- R3: a class with no name reads nothing and is not `Reconfigurable`.
- R4: the derived prefix matches the published table.
- `reconfigure` accepts the declared config class, whichever door built the instance, and rejects a foreign one.

**Construction contract** ([API conventions](https://grelmicro.grel.info/architecture/api-conventions/)):

- Every class with a matching `*Config` exposes `from_config` and no `config=` parameter. This audit was run by hand during 0.39.0 and found two outliers.
- Both construction doors leave the instance with the same attributes, which is what makes the `__new__` plus `_setup` pattern safe. Also run by hand.

## Why this is the highest-value test work available

It is not more tests, it is tests of a different shape. A per-class unit test cannot catch a rule that silently stops holding for the ninth member of a family, and that is what every defect above was. A sweep also covers implementations that do not exist yet, which is the only kind of test that protects a future contributor from a rule they never read.

It also converts the architecture docs from prose into something enforced. Three times during the 0.39.0 review a published contract turned out to be false in the code. Each was found by a human reading. A sweep makes that a CI failure instead.

## Done means

- A contract sweep per family, each refusing to pass on an empty scan, in the style of `_MIN_ADAPTERS`.
- Each rule in `docs/architecture/config.md` and `docs/architecture/plugins.md` either has a test naming it, or the doc says it is unenforced and why.
- Adding a new adapter or pattern that breaks a rule fails CI without anyone remembering to write a test for it.

## Architecture constraints

Settled decisions. Do not regress these while implementing.

**A sweep must fail on an empty scan.** `test_the_sweep_finds_adapters` exists because a scan that silently finds nothing passes forever. Every new sweep carries the same floor.

**Do not weaken a contract to make a test pass.** If an implementation cannot meet a rule, either fix the implementation or amend the doc with the reason, matching the Settled tables. Two of the 0.39.0 findings were resolved by amending the doc because the code was right, and that is a legitimate outcome, but it is a decision, not a shortcut.

**Exact-name lookup, never a prefix sweep**, wherever a test inspects environment variables. Kubernetes injects `{SVCNAME}_SERVICE_HOST` into every pod, which is why the shipped warning matches only declared field names.

**Do not fold this into the coverage gate.** Coverage is already 100% and did not catch any of these. Conformance is a separate axis and the point of this issue is that the two are not substitutes.

grelinfo:
Closes #755.

Three contract sweeps, each discovering its family by walking the package
instead of by a hand-written list, and each refusing to pass on an empty scan.

## Why discovery rather than a list

Every defect this issue was opened for was a rule that held for a family while
only one member was tested. The memory circuit breaker ran every algorithm as
consecutive-count while the other three adapters refused, and a list-based test
would have kept passing because nobody added it to the list.

`tests/resilience/test_ratelimiter_unknown_kind.py` had the right idea and the
wrong shape: it hand-listed four adapters, so it covered the family it was
written against and nothing added since. It is replaced here by a sweep that
finds eight and would find the ninth.

## What is enforced

**`tests/test_backend_contracts.py`** over every class declaring
`CircuitBreakerBackend` or `RateLimiterBackend` (8 today):

- `bind` refuses an unsupported kind with `NotImplementedError` naming it
- `bind` decides before touching the provider client, proved with a stub that
  raises on any attribute access, so an unsupported kind never needs a
  connection
- `scope` is declared and is one of `process`, `host`, `cluster`

**`tests/test_config_contracts.py`** over every `Reconfigurable` (15 today):

- R3: a nameless object is not `Reconfigurable`, pinned on `TTLCache`, which is
  the example the rule is written around
- R4: the derived prefix matches the published table, and a named instance
  falls back to the kind address (the row that silently stopped working in
  0.38)
- The config class `reconfigure` accepts is declared, and is frozen, so a
  reload swaps the object instead of mutating one an in-flight call already
  read

**`tests/test_construction_contracts.py`** over every class with `from_config`
(25 today):

- No `config=` keyword survives alongside it
- `from_config` never takes a parameter the config already carries, so no
  setting has two doors
- R8: it reads no environment variable and does not register for live reload,
  which would let a mounted file overwrite a config passed in code

## The rules I had to correct while writing them

Two of my first assertions encoded rules that do not exist, and the sweep
failing is what showed it.

*"`from_config` takes exactly one positional config"* is wrong. Patterns take
`(name, config)` because R3 makes the name the instance identity, and `Outbox`
also takes a `source`, which is the backend rather than a setting. The rule
that actually holds is the one now tested: `from_config` must not take a
parameter the config already carries. That is what makes a second door.

*"A `Reconfigurable` takes a `name`"* is wrong for components. `Tasks` is the
default instance and addresses itself through the kind prefix. R3 is about a
nameless object having no address, so the test now pins the case the rule is
written around instead of guessing at a signature.

## Verified by reintroducing the bugs

Each sweep was checked against the defect it exists to catch, not just for
being green:

- Removing the kind check from `MemoryCircuitBreakerAdapter.bind`, which is the
  exact 0.39.0 defect, fails the backend sweep
- Making `RedisRateLimiterAdapter.bind` read `self._provider.client` before
  dispatching fails it too
- Adding a `seconds=` parameter to `Timeout.from_config`, duplicating a config
  field, is caught by the duplicate-door rule

The sweeps also refused to hide their own gaps. An early version skipped seven
of twenty-five classes because `inspect.get_annotations(eval_str=True)` failed
on siblings imported under `TYPE_CHECKING`, and two more because a config union
is a forward reference served lazily by module `__getattr__`. A skip there
would have meant silently not checking a third of the family, which is the
failure mode this issue is about, so both are resolved rather than skipped.
There are now zero skips.

## Verification

Full `pre-commit run --all-files` passes: all 17 hooks, including the
integration tier and the 100% coverage gate.


grelinfo:
Both reviews triaged. Copilot raised 4, my own pass raised 8. All valid, all fixed.

**Copilot and I found the same defect independently**: the derived module name kept `.__init__` for package files. Copilot's diagnosis was that `pkg.__init__` "is not importable" and the sweep would error or skip. It is importable, which is what makes it worse: it re-executes the package under a second `sys.modules` name and hands back a class object that `is not` the public one. So the sweep was validating a duplicate nobody imports, and re-running import-time side effects. Fixed by naming a package after its directory. Its fourth comment, a missing comma, is applied.

**Two of my tests could not fail. That is worse than not having them**, and both were found by re-injecting the bug rather than by reading:

*The reload test.* `_reconfigurables` is a `WeakSet`, and the test called `from_config(...)` without binding the result, so each instance was collected before the count was read. `after == before` held whether or not registration happened. Proved by adding `_track_reconfigure` to `Timeout.from_config`: the test still passed. Now the instances are bound and the assertion is identity, `instance not in reconfigurable_instances()`, and the same injection fails it.

*The R4 prefix test.* It was parametrized over the published table but only ever called `env_prefixes(segment, ...)` with the test's own literal, so all eleven cases asserted that f-strings work. Every class passes its segment as a hard-coded literal at its own call site, so renaming `env_prefixes("TIMEOUT", ...)` would have broken `GREL_TIMEOUT_*` for every deployment with the sweep still green. It now reads the literal out of each class's AST and checks it against the table. Renaming it to `"TIMEOUTS"` fails the sweep.

**A third was a tautology**: `test_every_reconfigurable_declares_its_config_type` checked `issubclass(cls, Reconfigurable)`, guaranteed by how discovery works, and `getattr(cls, "config")`, a property on the base. Both branches unreachable. Deleted, since the sibling test does the real check.

**The R8 check grepped three strings** and missed `os.getenv`, `env_load_default()`, and `resolve_config_from_mapping`. A `from_config` calling `env_load_default()` passed while violating the rule. It now walks the AST for any name that reaches the environment, and that injection fails it.

**The duplicate-door rule checked one union arm.** `RateLimiter` resolved to `SlidingWindowConfig` alone, so a `capacity=` parameter, a `TokenBucketConfig` field, would have sailed through the exact rule meant to catch it. Now every arm is resolved and the parameters are intersected against the union of their fields.

**The floors had slack** (20 against 25 doors, 12 against 14). Up to five could drop out of discovery silently, which is a live path: `_discover_from_config` reads `node.body`, so moving `from_config` to a shared base would erode the sweep with everything green. Pinned to the exact counts, as the backend floor already was.

Also fixed: direct `os.environ` mutation that unset the variable for later tests under `xdist`, and a `monkeypatch.setenv` in a pure source-scanning test that did nothing.

Full `pre-commit run --all-files` green, including the integration tier and the 100% coverage gate.

PR #764, opened 2026-08-18
✅ Record real defaults in the API snapshot and share the sweep helpers

grelinfo:
Two things, both from auditing what #762 shipped rather than from new work.

## The snapshot was blind to changed defaults, on a false premise

`tests/test_public_api.py` blanked every default to `...`, justified in its own
docstring:

> Annotation reprs and default values vary across Python and Pydantic versions
> and would make the guard flaky across the supported matrix, so they are
> deliberately excluded. This catches a parameter rename, removal, reorder, or
> required-to-optional flip, but not a default-value change (that one is left to
> review and the changelog).

That premise is false, and it was never measured. I captured every default on
the public surface and compared across the whole supported matrix:

**All 392 defaults repr identically on 3.12, 3.13 and 3.14.** Exactly one value
varies, the memory address in a single `object()` sentinel shared by the
`Fallback` doors, which one regex normalises. Rebuilding the entire surface on
all three interpreters now produces byte-identical output.

So defaults are recorded. `Retry(attempts=3)` becoming `attempts=5` is a
behavioural break for every caller who never passed the argument, and it used to
be invisible here. Annotations stay stripped: those genuinely do vary.

### Why this closed the question of adopting griffe

#758 proposed replacing this snapshot with `griffe check`. Testing both against
real mutations of this repo:

| Change | snapshot before | griffe | snapshot now |
|---|---|---|---|
| Literal default `5.0` -> `2.0` | missed | caught | caught |
| Default via constant `_KEYED_MAXSIZE 1024` -> `16` | missed | **missed** | **caught** |
| Name dropped from `__all__`, import kept | caught | **missed, exit 0** | caught |
| Lazy `__getattr__` export dropped (real `AttributeError`) | caught | **missed, exit 0** | caught |
| Removal, rename, reorder, required-ness | caught | caught | caught |
| Return type changed | missed | missed (its own source: `# TODO`) | missed |

Griffe compares default *expressions*, so a constant it cannot resolve looks
unchanged. The snapshot resolves real values, so it now catches strictly more on
the one axis griffe was going to be added for. And griffe is static, so it exits
0 on both `__all__` shrinkage and a dropped lazy export, which this package's
`__getattr__`-assembled surface makes a live risk rather than a theoretical one.

No dependency is added. #758 gets updated separately.

## Hardening the sweeps from #762

- `module_name` was copy-pasted into three files. The `pkg.__init__` defect that
  made a sweep validate a duplicate class object had to be fixed in all three at
  once, which is the argument against the duplication. Now shared in
  `tests/_contract_support.py`.
- The R8 check only read `from_config`'s own body, so moving an environment read
  one function away passed it. Verified with an injected helper. It now follows
  module-level callees transitively.
  Deliberately not methods: following `_setup`, which both construction doors
  share, reported a prefix-building path `from_config` never takes.
- The backend discoverer did not handle `ast.Subscript` while its sibling did,
  and the parametrize `ids` had an off-by-one `rsplit`.

## Verification

Every claim above is a reproduction, not a reading. The two default-change rows
and the two griffe-misses rows were each produced by mutating one thing and
running both tools.

coverage gate.


grelinfo:
Both reviews triaged. Copilot raised 2, my own pass raised 7. Eight fixed, one rejected with evidence.

**Two findings undermined claims this PR makes, and both are now honest.**

*An interpreter-derived default slipped through the measurement.* `PickleSerializer(protocol=...)` defaults to `pickle.HIGHEST_PROTOCOL`, not a literal. My "all 392 defaults repr identically on 3.12, 3.13 and 3.14" measurement passed only because that constant is 5 on all three. A future CPython raising it fails this snapshot.

I kept recording the resolved value, because that failure is *correct*: the effective default really does change, and `PickleSerializer()` really does behave differently. What was missing is the explanation, so `test_interpreter_derived_defaults_are_still_current` now names the constant, and the docstring says so. A reviewer seeing `protocol=5` become `protocol=6` learns why rather than guessing.

*The "a changed default fails CI" claim was too broad.* A `default_factory` reprs as `<factory>`, so `RetryConfig.backoff`, the `worker` fields and the `headers` maps are still opaque. Changing what a factory returns passes silently. Both the module docstring and the changelog now say this instead of overstating.

**A latent bug in the shared helper.** `module_name` returned `grelmicro..` for the package's own `__init__.py`, because `Path(".").as_posix()` is `"."` and not `""`, so the fallback was dead code. Harmless today only because nothing is declared there. The first `Reconfigurable` or `from_config` class added to the top-level `__init__.py` would have made all three sweeps skip it with `ModuleNotFoundError` instead of checking it.

**The R8 callee walk followed too much.** It used every name mentioned, so a module-level `def timeout(...)` would be followed merely because a compliant `from_config` did `config.timeout`, and the door could then fail on a body it never runs. It now follows call targets only.

**The consolidation was half-applied**, which is the exact thing the helper module's docstring says it exists to prevent: `base_names` was still re-implemented inline in the config sweep, and `source_files` was added and never used. Both fixed.

**Also**: the changelog entry was a five-sentence paragraph against the one-line house style and linked #758, a different subject. Trimmed and repointed at this PR. Copilot's stale-docstring catch is applied.

**Rejected, with evidence**: Copilot flagged the attribute docstring after `_ADDRESS` as a Ruff B018 violation. `ruff check --select B018` passes clean, and attribute docstrings are this project's documented convention, used 16 times in `grelmicro/_config.py` alone.

**One claim in my own PR body was wrong** and I am correcting it rather than quietly dropping it: I described the `ids` change as fixing an off-by-one `rsplit`. It was not a fix. `f"{module}.{name}".rsplit(".", 2)[-1]` already evaluates to `name` for every path, so old and new ids are byte-identical. The change is cosmetic.

PR #765, opened 2026-08-18
✅ Record what a default factory returns and scrub numbers from errors

grelinfo:
Closes the three gaps #764 documented instead of fixing.

## A changed default factory was still invisible

`RetryConfig.backoff` picks the retry strategy for every caller who never
passes one, and it comes from a `default_factory`. The signature only carries
pydantic's opaque `<factory>` sentinel, so swapping `ExponentialBackoff` for
`ConstantBackoff` passed the guard silently.

The factory is now looked up on the model and called. Recorded:

```
RetryConfig(*, attempts=3, ..., backoff=ExponentialBackoff(kind='exponential',
    base_delay=0.1, max_delay=30.0, jitter='full'))
```

It is called **twice** and recorded only when both calls agree, so a `worker`
field minting a fresh identifier stays `<factory>` rather than writing a random
value into the snapshot. That distinction is the reason this is safe to do at
all.

Verified: swapping the default backoff strategy now fails the guard, and the
whole surface still rebuilds byte-identical on 3.12, 3.13 and 3.14.

## R8 could be bypassed across a module boundary

The check followed helpers the module defines, which closed the same-module
case. Putting the environment read in a neighbouring grelmicro module and
importing it still passed. It now follows first-party imported functions too,
stopping at the package boundary rather than descending into pydantic or the
stdlib. Verified by planting exactly that bypass.

## Errors scrubbed strings but not numbers

`_scrub` only removed string inputs, so a validator naming a rejected number
leaked it. A number read from the environment is a value from the environment,
and the reason to withhold it does not depend on its type. Booleans stay
excluded, since removing the words `True` and `False` would garble messages
that legitimately use them, and constraint messages are untouched as before.

## Verification

Each fix was checked by reintroducing the defect it closes, not by being green:
swapping the backoff factory, planting a cross-module env read, and scrubbing a
numeric canary.

coverage gate.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -23,7 +23,7 @@
 
 ### Added
 
-* ✅ The public API snapshot records default values, so a changed default fails CI. A `default_factory` still reprs opaquely. ([#764](https://github.com/grelinfo/grelmicro/pull/764))
+* ✅ The public API snapshot records default values, including what a `default_factory` returns, so a changed default fails CI. ([#764](https://github.com/grelinfo/grelmicro/pull/764))
 * ✅ Three contract sweeps enforce what the architecture docs publish, discovered by walking the package rather than by a hand-written list. `tests/test_backend_contracts.py` covers the `bind` contract over every backend adapter, `tests/test_config_contracts.py` covers R3, R4 and the reload rules over every `Reconfigurable`, and `tests/test_construction_contracts.py` covers the declarative door over every class with `from_config`. Each refuses to pass on an empty scan, so an adapter or pattern added later is covered without anyone remembering. ([#755](https://github.com/grelinfo/grelmicro/issues/755))
 * 📝 An [Errors](reference/errors.md) reference page. `SettingsValidationError` is now the one configuration error, and it had no documented home.
 * ✅ The public API snapshot covers `grelmicro.outbox` and `grelmicro.types`. Both were documented in the API reference while their exports went unguarded, so a rename could have shipped unnoticed. A test now reads the reference pages, so documenting a module guards it.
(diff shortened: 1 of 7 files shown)


grelinfo:
Both reviews triaged. Copilot raised 2, my own pass raised 7. All fixed.

**Copilot caught a claim in this PR that was false.** It said the numeric scrubbing the description promises was not in the code. It was right, and I had verified the change working earlier in the session, which made it worse rather than better: one of my `git checkout -- grelmicro/` cleanups after a bypass test reverted `grelmicro/errors.py` before the commit. `git show d84a35c3 --stat` confirms `errors.py` was never in it. The PR body described behaviour that did not ship.

Reapplied, and this time with `test_scrub_removes_a_rejected_number_too` so it cannot silently vanish again. That test is the actual fix: the code change had already existed once and was lost precisely because nothing held it.

**Four findings were the same mistake in different places: I closed the case I tested and left the neighbour open.**

- `_factory_default` was evaluated twice in one ternary, so a factory stable across the first pair of calls but not the next rendered as `v=`, a blank default. It also ran the factory four times per parameter, not the two the docstring claimed. Now computed once via a walrus. Verified: exactly 2 calls.
- The factory result skipped the `_ADDRESS` normalisation the literal path applies, so a factory returning a shared object would have written a memory address into the committed snapshot and failed every CI run after. Verified: now renders `<Shared object at 0x...>`.
- The cross-module walk was one hop. `functions` and `module` stayed pinned to the door's module, so a neighbour that delegated once more was invisible. It now carries `(source, owning_module)` pairs. Verified by planting a two-hop bypass, which is caught, alongside the one-hop and same-module cases which still are.
- `inspect.isfunction` rejected anything decorated, and `_build_settings_cls` is exactly that shape: an `lru_cache`d first-party helper that builds an environment-reading settings class. Now unwrapped.

**Two more:** an aliased field was matched by name only, so `Field(default_factory=list, alias="cee")` silently stayed `<factory>`, reopening the blind spot this PR closes. And `call_targets` returned attribute names, so `self._clock.monotonic()` would have resolved `monotonic` against the module namespace and judged a door on an unrelated function's body. Both fixed, and both bypass tests re-verified after narrowing.

PR #766, opened 2026-08-18
✅ Guard return types, the two-door invariant, and name validation

grelinfo:
Closes the three findings left open at the end of the contract work, including
two rules #755 listed and I closed it without.

## #755 shipped 8 of its 10 rules

**The two-door invariant was never implemented.** `from_config` skips `__init__`
entirely through `__new__` plus `_setup`, which is only safe while both doors
leave the instance in the same shape. #755 records it as audited by hand during
0.39.0 and asked for a sweep. There wasn't one.

Writing it turned a guess into a fact. My first assertion, "no difference", was
false: every pattern differs by exactly `_env_prefix`, and that difference *is*
R8 in structural form, since the reload address is what the declarative door
deliberately omits. So the test now asserts which attribute differs, not that
none does. "No difference" would have been wrong, and "some difference" would
let a genuinely missing attribute through. Verified by deleting `_scopes` in
`from_config`: caught.

**R7's remaining rows** are covered by `test_env_load.py` and
`test_env_cross_arm.py`, which existed already. They are not swept over the
family, and I am not claiming otherwise here.

## Return types are now guarded, which nothing else does

`Lock.from_config(...) -> Self` becoming `-> None` passed every guard. Griffe
does not catch it either, despite its docs: `_returns_are_compatible` ends in a
`TODO` and returns `True`. `cargo-semver-checks` declines it too.

Measured the same way defaults were, rather than assumed: **all 145 return
annotations on the public surface render identically on 3.12, 3.13 and 3.14.**
So they are recorded. Parameter annotations stay stripped, since those carry
`Annotated[..., Doc(...)]` payloads that genuinely do vary.

## A rejected name raises the one error

`Lock("bad name")` and `env_segment("123x")` raised a bare `ValueError`, so both
escaped `except SettingsValidationError`. Flagged during the #760 audit as
needing a deliberate ruling, which never happened. Making it now:

**A name is code, not a value.** R3 makes an instance name the address the
environment writes to, and R6 keeps structure in code. So the name is still
repeated in the message: it is a literal the caller wrote, not something read
from a variable, and the message is useless without it. R7's never-echo rule
governs environment values and does not reach it.

But the *type* becomes `SettingsValidationError`, because a caller should not
have to know whether their bad input counted as configuration or as identity.
It subclasses `ValueError`, so every existing `except` keeps working. The ruling
is written into `docs/architecture/config.md` so it is not re-derived.

## Verification

Each change was checked by reintroducing the defect: deleting an attribute in
`from_config`, changing a public return type, and constructing with a bad name.

coverage gate.


grelinfo:
Both reviews triaged. Copilot raised 3, my own pass raised 11. All fixed.

**Both reviewers independently flagged the same gap, and they were right: the two-door sweep was a hand-written list of four while the file's own discovery finds twenty-five.** That contradicts the file's stated premise. I hand-listed it because constructing arbitrary classes needs their required arguments.

The fix is to stop constructing. It now reads the source and compares the `self.<attr>` assignments in `__init__` against those in `_setup`, so it covers **20 classes** instead of 4, with 5 correctly skipped for not using that door at all. Verified by planting an `__init__`-only attribute on `Timeout`: caught.

**The name ruling was applied to 3 of ~11 sites**, which is the same mistake this session keeps producing: I fixed the case I tested. `PostgresLockAdapter(table_name="1bad")` and a Redis `prefix` that cannot survive a cluster still raised bare `ValueError`. Now consistent across all eight adapter modules plus the Redis prefix check.

**The change had zero test coverage.** Reverting `raise SettingsValidationError(msg)` back to `raise ValueError(msg)` passed the entire suite at the 100% gate, because every call-site test asserted the supertype. Two tests added, and reverting the fix now fails both.

**The snapshot baked private paths and PEP 563 differences into the public surface.** `str(annotation)` recorded
`-> <class 'grelmicro.cache.cached._CachedDecorator'>`, and mixed `-> Self` with `-> typing.Self` depending on whether the defining module used `from __future__ import annotations`. Either would fail the guard on a rename that changes nothing public. Now rendered through `inspect.formatannotation` with the module qualifier stripped, so a return type is recorded by name.

**Docs**: the breaking entry was filed under `### Added` and is now under `### Breaking`; `docs/architecture/config.md` said the rejected value is never echoed "with no exceptions" 24 lines before carving one out, so it now distinguishes a rejected *value* from a rejected *name*; the ruling gained a row in the decisions table alongside its two siblings; and `docs/migration.md` gained a symptom row and prose, which it was missing entirely. Two stale `Raises:` blocks fixed.

PR #767, opened 2026-08-18
👷 Cut the changelog and run the preflight with one command

### LINKED ISSUE #757: 👷 Define and automate the release procedure end to end
Cutting 0.39.0 surfaced a split: the *automated* half of releasing is in good shape, and the *human* half in front of it is undocumented. Someone who is not the author has to reconstruct it, which is exactly the wrong thing to leave in place while [looking for a co-maintainer](https://github.com/grelinfo/grelmicro/labels/help%20wanted).

## What already works, and should not be rebuilt

`.github/workflows/release.yml` is solid and follows current practice:

- `permissions: {}` at the top, with each job granting only what it needs.
- Every action pinned by commit SHA, not by tag.
- PyPI Trusted Publishing over OIDC (`id-token: write`), no long-lived token.
- Build provenance attested, then **verified before publish**, not after.
- The full test tier including the matrix gates the publish job.
- A `workflow_dispatch` path with a `dry_run` input that validates a version before its tag exists.

`just release-check`, `just release-notes`, and `just verify-release` are all good, and `verify-release` genuinely re-downloads from PyPI and checks the attestation against the file digest rather than trusting the workflow's own report.

## The gap

Nothing documents the sequence a human performs *before* the tag. Cutting 0.39.0 required working out four things that are written down nowhere:

1. **Move `## Unreleased` to `## X.Y.Z - <date>` in the changelog.** `tools/changelog.py check` enforces it, so it fails loudly, but no page says to do it. It is discoverable only by triggering the failure.
2. **The changelog bump needs its own pull request and must merge first.** `release-check` refuses unless `HEAD` equals `origin/main`, and `main` is protected, so the ordering is forced but unstated.
3. **The tagging step.** `gh release create` appears nowhere in the repo. The exact invocation, including `--target main` and `--notes-file`, has to be inferred from what triggers the workflow.
4. **How `release-notes` connects to the release body.** It is described as "what the GitHub Release body should hold", with nothing linking it to the command that creates the release.

The version rule itself *is* documented (`0.x.0` is the breaking position, in "About grelmicro versions"). It was still faster to derive it by counting breaking sections across 75 changelog entries than to find that paragraph, which suggests it belongs next to the procedure rather than only in the versioning essay.

## What to do

**Document the sequence.** A numbered "Cutting a release" section in `CONTRIBUTING.md`, next to "Before cutting a release", with every command in order and the ordering constraint stated. Short enough that it is followed rather than skimmed.

**Automate the mechanical step.** `tools/changelog.py` has `check` and `notes`. A `release <version>` subcommand that rewrites `## Unreleased` to `## <version> - <today>` removes the one step that is currently manual, easy to get wrong, and already machine-checked. The checker and the writer should be the same tool so they cannot disagree.

**Consider a single entry point.** A `just release <version>` that runs the changelog rewrite, then `release-check`, then prints the exact `gh release create` command rather than running it. Stopping short of tagging is deliberate: the tag is immutable and should stay a conscious act, but everything before it can be one command.

**Audit the rest against the same bar.** The workflow is in good shape, so this is confirmation rather than expected findings: verify the dry-run path still works, confirm no job has broader permissions than it needs, and check that a failed publish leaves nothing half-done.

## Why it matters more than it looks

A tag is immutable. A failure found after tagging burns the version and forces a patch bump, which has already happened once: 0.34.0 passed every other check and failed its release on Python 3.14. The preflight exists because of that, and it works. But a preflight only helps if the person releasing knows to run it, and right now that knowledge is not written down.

## Architecture constraints

Settled decisions. Do not regress these.

**The tag stays a conscious act.** Automate everything up to it. Do not add a path that tags and publishes from one command without a human deciding, because the result cannot be undone.

**`release-check` runs from `origin/main`, never a branch.** It refuses otherwise on purpose: anything else is not what would be tagged.

**Provenance is verified before publishing, not after.** That ordering is what the SLSA Build Level 2 claim in the README rests on.

**Keep actions pinned by SHA** and keep `permissions: {}` as the default. A convenience refactor must not widen either.

**One source of truth for the changelog format.** If a writer is added, the existing checker validates its output rather than duplicating the rules.

grelinfo:
Closes #757.

The automated half of releasing was already in good shape. The human half in
front of it was undocumented, and cutting 0.39.0 meant reconstructing four
things written down nowhere. This closes that half without touching the part
that works.

## One command for everything ahead of the tag

```bash
just release 0.40.0
```

It cuts the changelog, runs the full preflight, and prints the
`gh release create` command rather than running it.

**It stops short of tagging on purpose.** A tag is immutable, and 0.34.0 already
burned a version by failing after one. Everything that can be checked is checked
by the time it prints, and the irreversible step stays a conscious act.

The first run stops after the rename and says why, which is the ordering
constraint that was previously unstated:

```
The changelog is cut but not merged. release-check runs from
origin/main, and main is protected, so this needs its own PR:
...
Then run 'just release 0.40.0' again from main.
```

## The writer lives with the checker

`tools/changelog.py` gained `release <version>`, which renames `## Unreleased`
to `## <version> - <today>`. That was the one manual step in the procedure, and
it was already machine-checked by `check`, so the failure was loud but the fix
was folklore.

Both now share one parser, so they cannot disagree about the format, and
`test_release_then_check_agree` holds them to it: whatever the writer emits, the
checker accepts. A separate writer could format a heading the checker rejects,
and that failure would land at release time.

It is idempotent, so re-running after a later step fails is safe, and it refuses
to rewrite a section that already carries a date, because a released section
describes something that shipped.

## Documented, not inferred

`CONTRIBUTING.md` gains a numbered "Cutting a release" section: pick the version
(`0.x.0` is the breaking position pre-1.0, stated next to the procedure rather
than only in the versioning essay), run `just release`, merge the changelog PR,
run it again, then tag.

## Audit

The issue asked for confirmation rather than findings, and that is what it is:

- `permissions: {}` at the top, least privilege per job.
- Every external action pinned by commit SHA. The only two unpinned `uses:` are
  local reusable workflows, which cannot be pinned.
- Ordering is attest, then **verify**, then publish. That is what the SLSA Build
  L2 claim rests on and it is intact.
- The `dry_run` path skips attestation and verification, so it validates a
  version before its tag exists without producing an unusable attestation.

No changes made to `release.yml`.

## Verification

`just release 0.40.0` was run end to end against a scratch copy of the
changelog: it cuts, the existing `check` accepts the output, `notes` still
prints the body, re-running is a no-op, and cutting an already-released version
is refused. 7 tests cover the writer.

coverage gate.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -484,6 +484,52 @@ cannot drift away from what CI runs.
 `just release-notes 0.33.1` prints that version's changelog section, which
 is what the GitHub Release body should hold.
 
+### Cutting a release
+
+Pick the version first. Pre-1.0, `0.x.0` is the breaking position, so a
+release with any `### Breaking` entry takes the next minor and everything
+else takes the next patch. The queued entries are under `## Unreleased` in
+`docs/changelog.md`.
+
+Then one command does everything ahead of the tag:
+
+```bash
+just release 0.40.0
+```
+
+It renames `## Unreleased` to `## 0.40.0 - <today>`, then runs the full
+preflight, then prints the tag command for you to run.
+
+The first run stops after the rename, because the changelog change has to
+be on `main` before it can be released: `release-check` refuses unless
+`HEAD` is `origin/main`, and `main` is protected. So the rename goes in its
+own pull request:
+
+```bash
+git checkout -b chore/release-0.40.0
+git commit -am "🔖 Prepare the 0.40.0 release"
+gh pr create --fill && gh pr merge --squash --admin
+git checkout main && git pull
+```
+
+Run `just release 0.40.0` again from `main`. This time it reaches the
+preflight, and on success prints:
+
+```bash
+just release-notes 0.40.0 > /tmp/notes-0.40.0.md
+gh release create 0.40.0 --target main \
+    --title 0.40.0 --notes-file /tmp/notes-0.40.0.md
+```
+
+Creating the tag is deliberately left to you. It is immutable, and a
+version that fails after tagging is burned. Everything that can be checked
+has been checked by this point.
+
+Pushing the tag starts `release.yml`, which runs the full matrix again,
+builds, attests provenance, verifies that provenance, and only then
+publishes to PyPI over OIDC. When it finishes, confirm the published
+artifacts with `just verify-release 0.40.0`.
+
 ### Verifying a published release
 
 The README advertises SLSA Build Level 2. That claim holds only while
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -507,7 +507,7 @@ own pull request:
 
 ```bash
 git checkout -b chore/release-0.40.0
-git commit -am "🔖 Prepare the 0.40.0 release"
+git commit -m "🔖 Prepare the 0.40.0 release" -- docs/changelog.md
 gh pr create --fill && gh pr merge --squash --admin
 git checkout main && git pull
 ```
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -24,6 +24,7 @@
 
 ### Added
 
+* 👷 `just release <version>` runs everything ahead of the tag: it cuts the changelog, runs the full preflight, and prints the `gh release create` command rather than running it. Creating the tag stays a conscious act, because it cannot be undone. CONTRIBUTING documents the sequence, which was previously discoverable only by triggering the failures. ([#757](https://github.com/grelinfo/grelmicro/issues/757))
 * ✅ The public API snapshot records default values, including what a `default_factory` returns and the return annotation, so a changed default or return type fails CI. ([#764](https://github.com/grelinfo/grelmicro/pull/764))
 * ✅ Three contract sweeps enforce what the architecture docs publish, discovered by walking the package rather than by a hand-written list. `tests/test_backend_contracts.py` covers the `bind` contract over every backend adapter, `tests/test_config_contracts.py` covers R3, R4 and the reload rules over every `Reconfigurable`, and `tests/test_construction_contracts.py` covers the declarative door over every class with `from_config`. Each refuses to pass on an empty scan, so an adapter or pattern added later is covered without anyone remembering. ([#755](https://github.com/grelinfo/grelmicro/issues/755))
 * 📝 An [Errors](reference/errors.md) reference page. `SettingsValidationError` is now the one configuration error, and it had no documented home.
--- a/justfile
+++ b/justfile
@@ -132,7 +132,7 @@ release version:
         echo "origin/main, and main is protected, so this needs its own PR:"
         echo
         echo "  git checkout -b chore/release-{{version}}"
-        echo "  git commit -am '🔖 Prepare the {{version}} release'"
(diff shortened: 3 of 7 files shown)


grelinfo:
Both reviews triaged. Copilot raised 2, my own pass raised 7. All fixed.

**The worst finding is a deadlock in the exact procedure this PR documents.** `release` refused any date but today, and `check` demanded today, so a changelog PR cut on the 17th and merged on the 18th had no way forward. Re-running said "pick the next version", which is wrong: nothing shipped. UTC midnight is 02:00 local here, so an evening cut hit it with no delay at all.

The distinction that matters is **tagged, not dated**. A tagged section describes what shipped and is never rewritten; an untagged one is still being prepared and is re-dated. Verified both directions.

**Two silent no-ops, and the second is the same bug as the first.** `release` rewrote with a literal `text.replace("## Unreleased\n", ...)` while `sections` matches `^## Unreleased\s*$`, so a heading with one trailing space reported "changelog cut", changed nothing, and failed two steps later at `check`. That directly contradicted this PR's own "one parser, so they cannot disagree" claim, since the write path did not use the parser at all.

Copilot then found the same shape in the fix I had just written: re-dating built its pattern from the date it read, so an undated `## 0.40.0` looked for `- None`, matched nothing, and returned 0. Both now rewrite through the parser's pattern and refuse when nothing matched.

**Three more that could reach a tag:**

- `just release v0.40.0` wrote `## v0.40.0`, which `check` then *accepted*, so a bad heading reached the printed `gh release create`. Versions are now validated against the bare form every tag here uses.
- The printed instructions said `git commit -am`, which sweeps every modified tracked file into a PR titled "Prepare the release" and squash-merges it to protected main with `--admin`. Since `just release` does not require a clean tree first, having other edits present is the normal case. Now scoped to `-- docs/changelog.md`.
- A section holding only `### Added` and no bullets passed both the writer and the checker, producing a release body that is a bare sub-heading.

**Also:** entries merged under `Unreleased` after the cut are now reported, since they ship inside the tag while the notes say nothing about them. And `pyproject.toml` ignored `INP001` for `tools/*` on the grounds that it is "not an importable package", which this PR made false. Removed, and ruff confirms the ignore was dead.

Nine tests now cover the writer, one per finding.

grelinfo:
CI caught a real bug that my local run structurally could not, and it is worth recording.

`test_release_refuses_to_rewrite_a_released_section` passed locally and failed in CI. The cause is not the test: `is_tagged` shells out to `git tag --list`, and CI checks out shallow, so **a version tagged months ago answers "not tagged"**. The guard that stops a released changelog section being rewritten disappeared in exactly the environment nobody watches, while passing on every developer machine.

`is_tagged` now fails closed. If the repository reports no tags at all, the question is unanswerable rather than answered "no", so a released section is treated as released. Losing the ability to re-date in a tagless clone is the right side to err on: the alternative silently rewrites history describing a shipped release.

The test now pins `is_tagged` rather than depending on the clone, and a new test covers the shallow case directly.

This is the third time in this PR that the same shape appeared: a check that answers confidently when it cannot actually tell. The literal `text.replace` reported "cut" while changing nothing, re-dating built a pattern from a date that was `None` and matched nothing, and `is_tagged` answered "no" when it had nothing to look at. All three now refuse instead of guessing.

CI is green.

PR #768, opened 2026-08-18
✅ Assert what the sweeps used to skip and cover the release tool

grelinfo:
Loose ends from this session's contract work, found by auditing what shipped
rather than by adding anything new.

## Five silent skips became assertions

The two-door sweep skipped `TTLCache` and the four providers because their
`from_config` does not use the `__new__` plus `_setup` door. That is true, but a
skip records nothing: read at a glance it looks like coverage.

Those doors build through the ordinary constructor, so the invariant holds by
construction, and the sweep now says so. **Zero skips across all five contract
suites.**

## The release tool's failure branches were untested

`tools/changelog.py` shipped in #767 with its refusal paths uncovered: every way
`check` can refuse, the `notes` miss, the command-line dispatch, and `_git_tags`
when git cannot answer at all.

That last one matters most. `is_tagged` fails closed when a clone reports no
tags, which is what stops a released section being rewritten in CI, and nothing
covered the path that makes it fail closed. A bug in release tooling costs a
burned version.

**`tools/` is deliberately not in `coverage.run.source`**, so no number here is
gated and none is claimed. Bringing it under the gate would pull in five other
untested scripts and belongs in its own change.

## Tests that could not fail

Three of the new tests asserted an exit code that a broken branch would have
produced anyway. Each is now pinned to the message that distinguishes it, and
each was checked by deleting the branch it covers:

- `main()` ends with an unguarded `return notes(...)`, so deleting the `check`
  dispatch left every test green while `check` silently fell through to `notes`.
- `check`'s undated branch returns 1, and so does the wrong-date branch below
  it, so deleting the first changed nothing.
- The two-door assertion accepted `"return cls" in source`, which admits
  `return cls._instances[name]` — an object that never ran `__init__`, the exact
  case the assertion exists to reject. Narrowed to `cls(`.

Also: the `_git_tags` test patched `changelog.subprocess`, which is the real
module, so it made every `subprocess.run` in the worker raise for the duration.
Now scoped to the module's own reference.

## Audited and clean

Reported because the checks ran, not because they found something:

- **22 leak-and-escape probes** across every pattern, both construction doors,
  and the non-pattern surfaces: no value echoed, nothing escaping
  `SettingsValidationError`.
- **Every sweep floor is exact**, not slack: 8 backends, 14 reconfigurables, 25
  doors, 12 env cases.
- **The subclass sweep still bites.** Reintroducing
  `ResilienceSettingsValidationError` fails it. My first attempt to prove this
  produced a module that could not import, which proved nothing, so it was
  redone with a valid subclass.
- No dead leftovers from the `BOTH_DOORS` rewrite, no unused imports, no stale
  docstrings.

## Verification

coverage gate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -446,6 +446,7 @@ sort = "-Cover"
 skip_covered = true
 exclude_also = [
     "if TYPE_CHECKING:",
+    "if __name__ == .__main__.:",
     "class .*\\bProtocol\\):",
     "assert_never\\(.*\\)",
     "pytest.fail\\(.*\\)",
--- a/tests/test_changelog_tool.py
+++ b/tests/test_changelog_tool.py
@@ -10,6 +10,7 @@ tag is immutable, so getting it wrong costs a version.
 """
 
 import datetime
+import sys
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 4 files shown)


grelinfo:
Both reviews triaged. Copilot raised 2, my own pass raised 4. Five fixed, one rejected with evidence.

**Three tests could not fail.** Each is now pinned to the message that distinguishes it, and each was checked by deleting the branch it covers:

- `main()` ends with an unguarded `return notes(...)`, so deleting the `check` dispatch left every test green while `check` silently fell through to `notes` and returned 1 anyway.
- `check`'s undated branch returns 1, and the wrong-date branch below it returns 1 for `None != today`, so deleting the first changed nothing.
- The two-door assertion accepted `"return cls" in source`, which admits `return cls._instances[name]` — an object that never ran `__init__`, the exact case it exists to reject.

**Copilot pushed that last one further and was right to.** Even narrowed to `"cls(" in source`, a substring matches inside a docstring or a comment. It is now parsed:

```
real cls(1) call        -> True
return cls._instances[1] -> False
return cls.from_client(1) -> False
"cls(1)" in a docstring  -> False
```

**Rejected, with evidence.** Copilot said `is_tagged` returns `False` when `_git_tags()` returns `None`, because "`not None` is False", and that the fail-closed test would therefore fail. `not None` is `True`. Verified by running it: `_git_tags() -> None`, `is_tagged() -> True`. The test passes and the behaviour is the intended fail-closed one.

**Also fixed:** the `_git_tags` test patched `changelog.subprocess`, which is the real module, so it made every `subprocess.run` in the worker raise for the duration. Now scoped to the module's own reference, and the non-zero-exit path is covered too.

**One claim removed rather than fixed.** The PR body said "now 95%". `tools/` is not in `coverage.run.source`, so nothing enforces that and it would regress silently. Bringing it under the gate would pull in five other untested scripts, so the body now says the situation plainly instead of quoting a number.

PR #769, opened 2026-08-18
🔖 Prepare the 0.40.0 release

grelinfo:
Cuts the changelog for 0.40.0 and closes the gaps a review of the whole `0.39.0..HEAD` range turned up.

## Release cut

- `## Unreleased` becomes `## 0.40.0 - 2026-08-18`. The section has a Breaking block, so this is a minor on the 0.x line, and `docs/migration.md` already ships a `## 0.40` section.
- Two bullets were missing their PR link.

## Fixes found by the review

- **A timezone abbreviation was sent to the wrong continent.** The new `did you mean` hint used difflib's default cutoff over the whole database, so `PST` and `JST` both suggested `MST` and `CEST` suggested `EST`. Since the release stopped echoing the rejected value, that hint is all the operator reads. A short name with no region gets none now, and city segments are matched, so `Zurich` reaches `Europe/Zurich`, which it could not before.
- **`RedisProviderConfigError` and `PostgresProviderConfigError` removed.** They were the last two per-module subclasses, contradicting the invariant `docs/architecture/config.md` states and this release's central claim.
- **`Match.exception()` and `Match.exception_cause()`** changed from `TypeError` to `ValueError` on a public API with no Breaking entry, no migration section, and a docstring still promising `TypeError`.
- **`PostgresOutboxAdapter`** raised a bare `ValueError` for a rejected table name while every other adapter raised `SettingsValidationError`, which the changelog claimed covered table names.
- **`cleanup_interval`** raised a bare `ValueError` and echoed the rejected value, in the same constructor whose `table_name` raised `SettingsValidationError`.
- **`GREL_SHIELD_MAX_RATE`** reported the failure under `GREL_SHIELD_{NAME}_MAX_RATE`, a variable that is not set.
- **The scrub backstop skipped custom error codes**, so a `PydanticCustomError` from a third-party config class was never scrubbed.
- **The reload loop** logged a `reconfigure()` rejection message verbatim while the handler beside it logged only the class name.
- The subclass sweep only scanned `*.errors` modules, which is why it passed with the two provider subclasses live. It walks every module now and refuses an empty scan.
- `docs/architecture/config.md` described a scrub that splits the value into fragments, which the code deliberately does not do, and `docs/migration.md` quoted a timezone message that no longer exists.

https://claude.ai/code/session_01Heu2pjcxZkz3Gm81Paif67


grelinfo:
Rejected: the changelog tool renames the heading by design, and every prior release commit has this shape (0.39.0 at `263ded39`). The next contributor recreates `## Unreleased` with their entry, which CONTRIBUTING already instructs. An empty section would also parse as a real section in `tools/changelog.py`.

PR #770, opened 2026-08-18
✨ Render every rejection as an RFC 9457 problem detail

### LINKED ISSUE #740: ✨ Render rejections as problem details
grelmicro has a good error taxonomy and no way to put it on the wire. `AdmissionError` already covers every "turned away" case (rate limiter, bulkhead, open circuit breaker, non-blocking lock), and `TimeoutError` covers the rest. A user catching those still has to invent a response body.

From [the roadmap](docs/roadmap.md), "The HTTP edge": *"problem-detail error responses."*

**What to ship**

An exception handler, installable on FastAPI, Starlette, and Litestar in one call, that renders grelmicro errors as `application/problem+json` per [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457.html), plus the `ProblemDetail` model itself so a user can return one directly.

**The point is the extra field, not the envelope.** A problem detail that restates the status line is worth nothing. grelmicro knows which rejection happened and what the client should do next:

| Error | Status | Carries |
|---|---|---|
| `RateLimitExceededError` | 429 | `Retry-After` from `retry_after`, plus the `RateLimit` fields |
| `CircuitBreakerOpenError` | 503 | Time until the breaker next allows a probe |
| `BulkheadFullError` | 503 | Nothing to wait on, so say so rather than inviting a retry |
| `TimeoutError` | 504 | The deadline that was exceeded |

Each gets a stable `type` URI, so a client can branch on it without parsing prose.

**To settle while implementing**

- Where the `type` URIs point. A docs page per error kind, or an opaque URN.
- Whether an unmapped `GrelmicroError` renders as 500 or stays unhandled.
- Whether the handler is opt-in per app or wired by `micro.install(app)`.
- Whether `detail` may carry the exception message, given a message can name a backend or a key. Redaction rules apply.

## Architecture constraints

**Path exclusion is spelled `exclude=`.** Signature `exclude: tuple[str, ...] = ()`, matched against `scope["path"]`, exact match unless the pattern ends with `*`, which matches as a prefix.

Not `skip=`: that word already means "a predicate that decides an outcome is not persisted" on `@cached` and `IdempotencyMiddleware`. Not `paths=`: that word already means suffix-matched drop targets on `ProbeFilter`. Every new ASGI middleware (#739, #740, #741, #742) takes the same word with the same matching, so a reader learns it once.

**A pre-built config goes through `from_config`.** If this ships a config class, `from_config` is the only door for a pre-built one and the constructor is the live lane. Do not add a `config=` keyword: it was removed from every class that had one.

**Read result tuples by name.** `RateLimitResult` and `CircuitBreakerSnapshot` may gain fields with defaults. Never unpack them positionally.

**Carry the information the error already has.** A problem detail that restates the status line is worth nothing. `RateLimitExceededError.retry_after` becomes `Retry-After`, an open breaker reports the time until it next admits a probe. Read those fields by name.

grelinfo:
Closes #740.

`AdmissionError` already covered every "turned away" case and had no way to reach the wire. Now every rejection answers the client as an [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457.html) `application/problem+json` body, carrying the field the client can act on.

## Wiring

`micro.install(app)` registers the handler. `problem_details=False` opts out.

| Framework | How it lands |
|---|---|
| FastAPI, Starlette | `app.add_exception_handler(cls, ...)` |
| Litestar | `app.exception_handlers[cls] = ...` |
| FastStream | nothing, no HTTP to answer on |
| Any ASGI app | `grelmicro.http.send_problem` |

Every framework resolves a handler through the raised class's MRO, so registering four base classes covers every present and future subclass. A rejection added in a later release needs no new registration.

The `Integration` protocol stays frozen. `install_problem_details(app)` is a **new optional module attribute**, feature-detected with `getattr`, which is the extension path `grelmicro/_discovery.py` already prescribes. A third-party integration written today keeps working and gains problem details the day it adds the attribute.

## The mapping

| Error | Status | `type` anchor | Carries |
|---|---|---|---|
| `RateLimitExceededError` | 429 | `rate-limit-exceeded` | `retry_after` |
| `CircuitBreakerError` | 503 | `circuit-breaker-open` | `retry_after` |
| `BulkheadFullError` | 503 | `bulkhead-full` | nothing to wait on |
| `WouldBlockError` | 503 | `lock-unavailable` | nothing to wait on |
| `AdmissionError` | 503 | `request-refused` | nothing to wait on |
| `DeadlineExceededError` | 504 | `deadline-exceeded` | `timeout` |
| `IdempotencyConflictError` | 422 | `idempotency-key-reused` | nothing |
| `IdempotencyWaitTimeoutError` | 409 | `idempotency-in-flight` | `retry_after` |

Every `type` URI dereferences to its own section on the docs page. An error grelmicro did not raise to turn a caller away stays unhandled, so a bug is still a `500`.

## Two errors gained the field that makes them worth rendering

A problem detail that restates the status line is worth nothing, and two errors could not carry anything.

- **`CircuitBreakerError.retry_after`**, plumbed through `CircuitBreakerSnapshot.retry_after` and computed by the strategy, because `opened_at` is on the strategy's own clock. Implemented in all four adapters: memory, Redis (Lua), Postgres (PL/pgSQL), SQLite. A `FORCED_OPEN` breaker reports `0.0`, so it never invites a retry that cannot succeed.
- **`DeadlineExceededError`**, raised by `Timeout` instead of a bare builtin. It subclasses `TimeoutError`, so `except TimeoutError` keeps working, and it carries `name` and `timeout`. The deadline travels with its scope, so a reconfigure mid-call cannot make the error report a deadline the block never ran under.

## Secure by default

- **The exception message never reaches the wire.** Every `detail` is written by grelmicro. `RateLimitExceededError` alone would have published the key it rejected, which is often a client address or a user id. A test asserts `str(exc)` appears in no rendered body, for every rejection.
- **`Cache-Control: no-store`** on every problem, so a shared cache cannot serve one client's `429` to another, or keep refusing after a dependency recovers.
- **`X-Content-Type-Options: nosniff`**, because `instance` reflects the request path back.
- No breaker name, bulkhead name, or rate limit key is published. Only the delay a client can act on.

## Breaking

- `Timeout` raises `DeadlineExceededError`. `except TimeoutError` keeps working.
- `IdempotencyMiddleware` answers with `application/problem+json` instead of `{"detail": ...}`. Its `400`, `409`, `413`, and `422` all change shape.
- The Postgres circuit breaker drops and recreates three stored functions on the next start, because their result columns grew. Postgres refuses `CREATE OR REPLACE FUNCTION` when the returned columns change, so `auto_migrate` drops first, under the advisory lock it already takes. An integration test installs the old signature and checks the upgrade comes up.

## Also

`document_idempotency(app)` publishes a `ProblemDetail` component and points each middleware response at it, so a generated client knows the body it will get.

## Verification

Full pre-commit chain green: ruff, ty, mypy, unit tier, **integration tier** (Redis and Postgres containers, including the migration test), 100% coverage, `mkdocs build --strict`. 4800+ tests.

https://claude.ai/code/session_01Heu2pjcxZkz3Gm81Paif67


grelinfo:
## Cross-framework parity, measured

The same rejection now has to produce the **same response byte for byte** on every framework grelmicro serves HTTP on, not merely the same status and shape.

| | FastAPI | Starlette | Litestar |
|---|---|---|---|
| status | 429 | 429 | 429 |
| `content-type` | `application/problem+json` | same | same |
| `cache-control` | `no-store` | same | same |
| `x-content-type-options` | `nosniff` | same | same |
| `retry-after` | `2` | same | same |
| `content-length` | 297 | 297 | 297 |
| body | reference | identical | identical |

`test_every_http_framework_answers_identically` compares status, the full header map, and the body bytes across all three. A client cannot tell which framework answered, so a service can move between them without its callers noticing.

The test is not vacuous. Giving Litestar a `; charset=utf-8` on its media type fails it:

```
>       assert headers == expected_headers, name
E       assert {'cache-contr...t=utf-8', ...} == {'cache-contr...': '297', ...}
```

FastStream is absent from the matrix because it serves no HTTP. `test_faststream_takes_no_problem_details` asserts the integration carries no `install_problem_details`, so `micro.install(app)` skips it by feature detection rather than by branching on the framework's name.

Broader coverage of this across every pattern and component is #771.

grelinfo:
## Code review findings, all seven evaluated

I reproduced each finding against the code before touching anything. All seven were real. Six are fixed here, the seventh is a design question I raised rather than decided.

### Fixed

**1. `install_problem_details` clobbered a handler you registered first.** Confirmed: registering `AdmissionError` before `micro.install(app)`, or passing `Litestar(exception_handlers=...)`, returned grelmicro's 503 instead of the user's 418 on both frameworks. That is the natural order (build the app, register handlers, wire grelmicro) and it is the order that lost. Both loops now skip a class that already has an entry. `test_a_handler_registered_before_install_wins` and `test_litestar_keeps_a_handler_passed_to_the_constructor` cover it, alongside the existing after-install test.

**2. `body_of` used `model_dump()` instead of JSON mode.** Confirmed: `Decimal("30.00")` as an extension member raised `TypeError` from inside the response. And `UUID` succeeded *only because orjson is installed* — with the stdlib fallback in `_json.py` the same page 500s. A page that works in dev and fails in prod is the worst shape of this bug. Now `model_dump(mode="json", ...)`.

**3. `_wait` rounded after the zero guard.** Confirmed: `retry_after=0.0004` passed `> 0`, rounded to `0.0`, and published `"retry_after": 0.0` with `Retry-After: 0` — the exact "retry now" the docstring says it never emits. A limiter reports a sub-millisecond wait at a window boundary under load. Rounds first now.

**4. `retry_after_of` assumed the extension was numeric.** Confirmed: a hand-built `ProblemDetail(retry_after="30")` through the public `send_problem` raised `TypeError: must be real number, not str`. A refusal must not fail while refusing. Non-numbers (and `bool`) now carry no header.

**5. The Postgres drop ran on every start, forever.** The sharpest finding. It dropped and recreated three functions on every restart of every replica, assigning new OIDs and invalidating every dependent cached plan cluster-wide — permanently, for a one-time migration. Now gated on the stored result signature via `pg_get_function_result`. `test_migrate_leaves_current_functions_alone` asserts the OIDs are unchanged across a restart, and I confirmed it fails against the old unconditional version:

```
At index 1 diff: {'proname': '..._cb_get_state', 'oid': 16416} != {..., 'oid': 16413}
```

The upgrade test still passes, so the migration still works when it is actually needed.

**7. The OpenAPI component could point at someone else's model.** Confirmed. `_merge_response` hard-coded `#/components/schemas/ProblemDetail` while `_add_problem_schema` only `setdefault`-ed it, so an app publishing its own `ProblemDetail` had the middleware's responses reference its shape. Now the component is compared before reuse, and grelmicro's is published beside it as `GrelmicroProblemDetail`. If both names are taken, the responses stay described and carry no body schema — naming the wrong shape is worse than naming none.

### Raised, not decided

**6. `Lock.acquire(timeout=...)` raises a bare `TimeoutError` and answers 500.** The finding is right and the docs sentence overclaimed. Two neighbouring failures on the same primitive answered very differently:

| Call | Raises | HTTP |
|---|---|---|
| `lock.acquire(blocking=False)` | `WouldBlockError` | `503` `lock-unavailable` |
| `lock.acquire(timeout=5.0)` | builtin `TimeoutError` | `500` |

Fixing this properly means changing `Lock`'s public error contract, and the right status, the right name, and whether it should merge with `WouldBlockError` are genuine design calls — not something to slip into a problem-details PR. I filed **#772** with the options and constraints, and made the docs precise here instead of leaving a promise the code does not keep. `Shield`'s bare `asyncio.timeout` is in that issue too.

### Not changed

The review also confirmed a set of things I had reasoned about but not proven: MRO dispatch order, the `Timeout.__aexit__` conversion not misconverting an inner unrelated `TimeoutError`, the Redis Lua `cool_down` write/`HDEL` paths, `{remaining}` brace safety in the PL/pgSQL, Litestar's `Response.render` not double-encoding bytes, and Starlette's `URL.path` matching `scope["path"]` so handler and middleware `instance` agree. Good to have those pinned.

Full chain green again: ruff, ty, mypy, unit, integration (real Redis and Postgres), 100% coverage, `mkdocs build --strict`. 4820 tests.

PR #782, opened 2026-08-20
🐛 Never let a caller's object raise out of a matcher or a validator

grelinfo:
## The bug

An object the caller supplies is caller code all the way down. `__class__`, `__name__`, `__instancecheck__`, `__str__`, `__repr__`, `__cause__`, `__bool__`, `__eq__`, `__hash__`: every one of them is read while grelmicro decides what a value is, and every one of them can raise.

Two places in the library cannot survive that, and both did it anyway.

**A matcher.** `Retry` calls it from inside `except Exception as exc:`, so anything raised there replaces the exception the caller is already handling. A broken predicate did not report itself, it destroyed the real error and reported the wrong one.

**A validator.** Pydantic converts `ValueError` and `AssertionError` and nothing else, so a `TypeError`, a `re.error`, or whatever a lazy proxy raised from `__class__` escaped `except SettingsValidationError` entirely. `Retry(when=...)`, `Fallback(when=...)` and a shield's `timeout_errors=` all classified their argument with a bare `isinstance`.

Worst of the set: **`SettingsValidationError` itself**. Deciding what to redact ran `isinstance` on the rejected value and rendered it, so an unbound `werkzeug.local.LocalProxy` replaced the error with its own and skipped the redaction that keeps a credential out of the message. That reproduces on a plain pydantic model, so it was never about `Match`.

Separately, the warning registry answered for the wrong predicate: the same recycled-`id()` defect [#781](https://github.com/grelinfo/grelmicro/pull/781) fixed for lock tokens. `_warned_predicates` was a `set[int]` of `id(predicate)` holding nothing alive, so an entry outlived its predicate and silently suppressed the next one allocated at that address, while the entries accumulated for every predicate ever built.

## The fix

**One home for the shape tests.** `grelmicro/_guards.py` holds `is_instance`, `is_class`, `is_subclass` and `type_name`, each total: a hostile value answers `False`, a genuine `KeyboardInterrupt` or `SystemExit` still propagates. `_match.py`, `retry.py`, `fallback.py`, `shield/_profile.py` and `errors.py` all use them instead of each keeping a copy.

**A matcher never raises, and neither does reporting one.** One guard at the single call entry point, total helpers for each step that reads caller code, and a total `_warn`, because `Logger.handle` runs filters unguarded and a request-context filter raising outside a request escaped from the warning itself.

**A warning names a predicate, it never prints one.** A `functools.partial` holds the arguments bound into it and an object predicate holds its attributes, so a `repr` put an API key into the log. A partial reads as the function it wraps, so two partials in one policy stay tellable apart. A literal keeps its `repr`, because `Match.result(200)` should read as itself.

**Every argument error is a `ValueError` at construction**, across `Match` and the three config validators. The operators are the deliberate exception: they return `NotImplemented`, which is Python's protocol and lets the other side answer with `__ror__`.

**One entry per predicate, for exactly as long as the predicate lives.** A `WeakSet` of the predicates themselves, with a bounded FIFO fallback by address for the ones a `WeakSet` cannot hold. Nothing enters the weak set that could not find itself again: a moving `__hash__` or an `__eq__` that disagrees with itself is asked before the set is touched, never after, so no entry is ever stranded in it.

## Tests

Every guard is covered from both sides: the hostile value is absorbed, and a genuine interrupt raised from the same dunder still propagates. The registry is measured across many eviction cycles rather than one. `grelmicro` is at 100% line and branch coverage, and a script re-checks every claim this changelog makes in one run.

Two tests I wrote during this PR **passed against the buggy code** and had to be replaced. The first forced CPython to reuse an address, which is not deterministic and actually masks the growth it was meant to expose. The second measured the registry across a single eviction cycle, which is exactly one fewer than the leak needed to show.

https://claude.ai/code/session_01BH2HHdveQ5RK19Wqg7EqTg


grelinfo:
Triage of the three suppressed comments.

**`_describe` and `str.__str__` — refuted.** `str.__str__` does coerce. CPython's `unicode_str` returns `unicode_result_unchanged`, which returns the same object only for an exact `str` and otherwise makes an exact-`str` copy. Reproduced against this branch with a `str` subclass whose `__format__` and `__str__` both raise:

```
str.__str__(Evil('x')) is exact str: True
_describe via __name__  -> 'NameIsEvil' exact str: True
_describe via __repr__  -> 'evil-repr'  exact str: True
```

Interpolating either result runs no caller code. `test_naming_a_predicate_returns_an_exact_str` covers it.

**`_message_of(None)` returning `"None"` — valid, fixed.** The parameter is now `BaseException`, and both message matchers read `outcome.exception` first and answer `False` when there is none. `_message_of` is never reached without an exception, so the string `"None"` can no longer be produced. `test_a_message_matcher_ignores_a_returned_value` covers it.

**`_coerce_bool` docstring — valid, fixed.** The sentence described the guard as it was before `Match.__call__` took over. Removed.

grelinfo:
Round seven, from an adversarial pass plus this review. Nine findings, all reproduced before being fixed.

**The warning path could raise out of a matcher.** Every caller-controlled dunder was guarded, but `_log.warning` was not, and `Logger.handle` runs filters unguarded. A request-context filter raising outside a request escaped from the very place a matcher reports a problem, replacing the exception being handled. The non-bool case was the worse one: an ordinary predicate returning `1`, not a hostile object. All warnings now go through a total `_warn`.

**A warning printed what a predicate holds.** `_describe` fell back to `repr`, so `functools.partial(check, api_key='sk-live-...')` and an object predicate holding a token both reached the log. Naming is now a name, never a `repr`. A literal keeps its `repr`, because `Match.result(200)` should read as itself.

**Two bare `isinstance` calls.** `_compile_pattern` and the operators skipped the guard the module has `_is_class` for. A lazy proxy raising from `__class__` escaped both. There is now one `_is_instance`, and `_is_class` is a call to it.

**A bytes pattern was accepted** and, now that the matcher is total, would have become a silent never-match. Refused at construction. Found here and by the adversarial pass independently.

**`Match.predicate(42)` built a policy that never engages.** Refused at construction.

**`Match.exception_message()` mutual exclusion raised `TypeError`,** the one argument error the PR had left inconsistent. Every `Match` argument error is a `ValueError` now.

**A predicate whose `__hash__` changes** never found its own `WeakSet` entry, so the set grew one entry per call and it warned every call, against both promises the registry makes. It is now recognized by address instead.

**Three changelog claims were wrong.** "Both used to build and then fail at match time" held for `contains=` only, both regex forms already raised at construction with the wrong error. The `__str__` case is reported zero times, not once. And `Match.predicate` now coercing its return was undocumented. All three corrected.

**Tests.** Two tests whose names promised a count never checked one, and one asserted `is not None`. They now count warnings through `caplog` and assert the policy label. The two hostile metaclasses crashed pytest's own reporter on any failure in that file, which would have hidden the real failure the day one broke, so they got a safe `__repr__`.

On the `__or__` signature: `other: Match` is deliberate. Typeshed keeps the narrow operand type on arithmetic dunders so a static checker flags `policy | 5` at check time, while `NotImplemented` gives the standard `TypeError` at runtime. Widening to `object` would silence the checker. Both `ty` and `mypy` accept `return NotImplemented` under `-> Match`.

`_match.py` is at 100% line and branch coverage, and a single script re-verifies every claim in the changelog: 45 checks, all passing.

grelinfo:
Three comments this round, all refuted, one with a test added so the answer is checked rather than argued.

**`str.__str__` and `str` subclasses (twice).** Same claim as the previous round, same answer: `str.__str__` does coerce. CPython's `unicode_str` returns `unicode_result_unchanged`, which returns the argument only when `PyUnicode_CheckExact` holds and otherwise makes an exact-`str` copy. Reproduced against this branch with a subclass whose `__format__` and `__str__` both raise, through `__name__`, through `__repr__`, and through `_describe_value`: every result is `type(...) is str` and interpolates without running caller code. `test_naming_a_predicate_returns_an_exact_str` and `test_a_literal_that_cannot_be_repred_still_names_the_policy` cover both helpers.

**`_warned_predicates` growing for a changing `__hash__`.** The excerpt stops one line above the part that decides it. `_already_warned` asks the address registry **first**:

```python
if id(predicate) in _warned_untrackable:
    return True
```

So the first call leaves one stray weak entry, the `discard` misses as you say, the predicate is recorded by address, and every later call is answered by that check without touching the `WeakSet` again. The growth is one entry, once, not one per call.

`test_a_predicate_whose_hash_changes_stays_bounded` now proves it over 200 calls: at most one weak entry, exactly one warning, and after 128 newer untrackable predicates evict it, the weak set is back to its starting size and the address registry is still at its bound. The address registry holds a strong reference deliberately, so the address it keys on cannot be handed to a different predicate while it is still remembered.

grelinfo:
Round nine, from a final pre-merge pass plus this review. Verdict came back MERGE with one line worth fixing first, and it was right.

**`type_name` was not total, in code this PR introduces.** It returned `str(type(value).__name__)` without the exact-`str` normalization `_describe` applies for exactly this reason, so a metaclass whose `__name__` is a `str` subclass with a hostile `__format__` survived the helper and blew up at the line that interpolated it, inside a validator. Both call sites are new code. Reproduced, fixed, and covered from both sides.

**The message I added yesterday was broken.** `type_name(type(item))` always renders `type`, because the item is itself a class: `timeout_errors entry type is not an Exception subclass`. There is now a `name_of` in `_guards`, which names a class by itself and anything else by its type, and the older `value.__name__` site uses it too.

**A shield narrowing that had no changelog line.** Classifying the value before pydantic sees it means a set, a `deque`, and a generator no longer slip through its lax tuple coercion. Tuples, lists, CSV, JSON and FQNs are unaffected. Now under `### Breaking`.

**The scrubber's traversal was not as total as its docstring.** A `dict` subclass whose `values()` raises, or a `list` subclass whose `__iter__` raises, still replaced the `SettingsValidationError`. The walk is guarded now, and a `str` subclass is read with `str.__str__` rather than by asking its `__str__`, so the value is redacted instead of silently skipped.

**One docstring overclaimed.** `_finds_itself` said nothing is ever stranded. A hash the program itself moves, by mutating the attribute it is built from between calls, is stable when asked and unstable later. That shape behaves identically on `main`, so it is not a regression, and the docstring now says what the gate actually promises.

Also on Copilot's finding: `timeout_errors=KeyboardInterrupt` was refused, `timeout_errors=(KeyboardInterrupt,)` was accepted and then never retried. All five routes agree now, with a test for each.

Verified: 4766 unit and 293 integration tests, 100% line and branch coverage, and three scripts that re-check every claim in the changelog end to end.

grelinfo:
All five comments are one finding, and all five are right. Reproduced before fixing:

```
Retry(when=hostile tuple)              ESCAPED RuntimeError: iter exploded
Retry(when=hostile list)               ESCAPED RuntimeError: iter exploded
Fallback(when=hostile tuple)           ESCAPED RuntimeError: iter exploded
Fallback(when=hostile list)            ESCAPED RuntimeError: iter exploded
shield timeout_errors=hostile tuple    ESCAPED RuntimeError: iter exploded
shield timeout_errors=hostile list     ESCAPED RuntimeError: iter exploded
```

`__iter__` is caller code like every other dunder this PR guards, and a container subclass writes its own. It is the same shape I had just closed in the error scrubber, one layer up, so it now uses the same answer: `items_of` in `grelmicro/_guards.py` walks a container once and answers `None` when it will not be walked. `None` falls through to the argument error the value was going to get, so there is one message rather than a new one:

```
Retry(when=hostile tuple)              SettingsValidationError
Fallback(when=hostile list)            SettingsValidationError
shield timeout_errors=hostile tuple    ValidationError
```

Walking once also removes a second read: `_coerce_when` used to iterate the container twice, in the `all(...)` guard and again in the comprehension, so a container that answered differently the second time was already reading inconsistently.

While there: `_coerce_to_match`'s own error message interpolated `type(value).__name__` directly, which a hostile metaclass raises from. It uses `type_name` now, like everywhere else.

Six tests, one per route, plus three on the helper. 4775 unit and 293 integration tests, 100% line and branch coverage.

